PR #48, opened 2017-01-05
State copy should clone dictionaries

joningold:
When copying a state, the visitCounts and turnIndices objects need to be cloned, so that later changes don't change their entries. Otherwise, you get strange effects, such as with the following ink source:

-> bug
=== bug
    Removing this line resolves the issue.
    {not knot:  
         // Including a line here resolves the issue
        -> knot ->
    - else: 
        This content should NEVER be seen.
    }
    -> END
    
=== knot
    This content SHOULD be seen.
    -> END

Note there may be other members of the State object which need cloning as well, but my comparison with the c# suggested this is all that's needed?

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -717,6 +717,15 @@ export class StoryState{
 
 		copy.previousContentObject = this.previousContentObject;
 		
+		copy._visitCounts = {};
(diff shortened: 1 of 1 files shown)


y-lohse:
Thanks, good catch. Pushing out a new version with the fix.

PR #51, opened 2017-01-22
Fix for ink functions called from game code

joningold:
The following ink source: 

=== function not_working() ===
    Text1.
    Text2.
    ~ return true

in conjunction with a call like 

story.EvaluateFunction("not_working");

from the javascript file running the game causes an error when the ~ return statement is encountered as it's unexpected.

Note the C# doesn't seem to copy this property explicitly, but it is correct and does the fix the issue.

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -717,6 +717,8 @@ export class StoryState{
 
 		copy.previousContentObject = this.previousContentObject;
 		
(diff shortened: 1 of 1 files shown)


y-lohse:
Aah, nice catch, thank you.

Does anyone needs this fix urgently in a published version? Otherwise I'll make it a part of the next major release, which shouldn't be too far away now.

PR #52, opened 2017-01-23
Core ink fix: callstack used for external function

joningold:
When running a function using "story.EvaluateFunction", the ink caches the old callstack and makes a new one until the function evalutation is complete. However, the VariablesState object holds a pointer to the callstack which it uses for creating temporary variables. 

This fix updates the VariablesState objects callstack when the function evaluation begins, and resets it once complete. 

(Matches main ink repo commit 7477026cf35ee06fa93eee194e01bf03fe94c9a8)

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -610,6 +610,11 @@ export class StoryState{
 		// Create a new base call stack element.
 		this.callStack = new CallStack(funcContainer);
 		this.callStack.currentElement.type = PushPopType.Function;
+		
+		// The variablesState object needs to track the callstack 
+		// we're using so temp variables are created in the right place.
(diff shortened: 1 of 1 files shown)


y-lohse:
Thanks! 
If I'm not mistaken, this change happened *after* ink 0.7.0 and is not officially out yet, right? If so I'll wait to merge this PR until it's part of a published ink version, or I'll end up being all confused :D

joningold:
Yeah, we fixed this the other day and I thought I'd do the PR so you had it
to hand. So merge at your leisure!

Cheers
Jon

y-lohse:
Looks like we merged that one in #53!

PR #53, opened 2017-03-07
Fixes to external function evaluation

joningold:
A couple of lines required to fix external function evaluation seem to have fallen out of the latest version. (However, you'd better check that they really did get missed by mistake rather than being taken out deliberately!) And apologies, but I don't have a decent test case -- if I run my local js ink project, I get the "expected" results with these fixes, but not without.

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -631,6 +631,8 @@ export class StoryState{
 		// Create a new base call stack element.
 		this.callStack = new CallStack(funcContainer);
 		this.callStack.currentElement.type = PushPopType.Function;
+		
+		this._variablesState._callStack = this.callStack;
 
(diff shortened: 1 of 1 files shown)


erkyrath:
Speaking of external functions... Lectrote has no way to load externals, and it never will (for security reasons). But fallback functions don't work. The engine throws:

> Missing function binding for external: 'extfunc' (ink fallbacks disabled) StoryException

Am I supposed to directly set `story.allowExternalFunctionFallbacks = true`? After creating the story instance, I mean.

joningold:
Yes; I've always done it by hacking the ink.js file directly but you need
to set that true.

Jon

y-lohse:
Thank you! I've merged it into the dev branch, release to follow.

PR #55, opened 2017-03-08
Bugfix: previous container list wasn't cleared

joningold:
The equivalent c# for this section is 

// First, find the previously open set of containers
			if( _prevContainerSet == null ) _prevContainerSet = new HashSet<Container> ();
			_prevContainerSet.Clear();

...with the if'd statement being there to avoid multiple allocations, but the Clear() being called every time. The JS was missing this clear, resulting in previous containers hanging around.

### DIFF
--- a/engine/Story.js
+++ b/engine/Story.js
@@ -446,7 +446,7 @@ export class Story extends InkObject{
 			return;
             
 		// First, find the previously open set of containers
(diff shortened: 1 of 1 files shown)


y-lohse:
Good catch, Thanks for taking the time to submit the bug, find the origin *and* send a patch :smiley: .
Merged in the dev branch as well.

PR #58, opened 2017-03-18
Fix function declaration inside nested statement

erbridge:
This is an error in strict mode, which is forced by Babel.

```
SyntaxError: Strict mode does not allow function declarations in a lexically nested statement.
```


y-lohse:
Thank you :tokyo_tower: 
I merged the PR in the dev branch, it'll be part of the next release.

erbridge:
Any idea when the next release will be?

y-lohse:
Yup, probably this week-end.

PR #68, opened 2017-06-12
Fix: logic returned int not IntValue

joningold:
This bug's been here for a while, but I think this code path wasn't being used (much?) until 0.7.3 introduced "normal" && and || for lists.


y-lohse:
Thank you, good catch. FYI I'll remove the space before the parentheses to stay consistent with the rest of the codebase, but I don't want to bother you editing the PR for that :). Merged into dev! (I should probably make dev the default branch)

PR #71, opened 2017-06-29
Tests

y-lohse:
Ok, it's well past time ink.js gets some tests.

The big challenge will be to keep the source ink files, the compiled ink files and the expected test results in sync. Sadly I don't see a way to automate that. Here's my planned approach:

- [x] Write one (or maybe a few) ink files which can be used to test loads of things, probably grouped in knots / stitches, and add this file to source control
- [x] Add the corresponding compiled JSON to source control, and ideally add a command to regenerate it
- [ ] Write the actual tests

And here's a rough list of things that need testing:

### Content

- [x] simple content
- [x] knots and stitches
- [x] diverts (to knots, stitches, local diverts, in lists, gathers, variables?)
- [x] glue
- [x] conditional text
- [x] print variables

### Simple lists

- [x] sequences
- [x] cycles
- [x] once only
- [ ] nested
- [x] with blanks

### Tags

- [x] regular tags
- [x] tags on choices
- [x] knot level tags
- [x] global tags
- [x] empty tags / tags that could evaluate to weird stuff

### Choices

- [x] choices and choice text control
- [x] conditional choices
- [x] sticky choices
- [x] fallback choices
- [ ] stuff that happens after choices (txt, diverts, etc)

### Game queries

- [x] choice count
- [x] turns since
- [x] read count

### Variables and logic

- [x] global variables of all types
- [x] all operations
- [x] if / else
- [x] temp variables
- [x] knots and stitches with params
- [x] constants
- [x] custom functions (with and without params, with and without return, inline and block calling)

### Flow control

- [x] tunnels
- [x] threading
- [ ] ...there's probably more to it

### Advanced lists

- [ ] Lots of things here, running out of time

### Integration

- [x] external functions
- [x] variable observers
- [x] calling ink functions
- [x] saving and loading
- [x] jumping to a scene
- [x] get / set variables
- [x] visit counts

I assume some of these are actually redundant because they get taken care of by the compiler. Having code coverage report should help narrow things down

- [x] Add code coverage reports


joethephish:
I was actually starting to wonder if there's a way we could roll one test system together that tests:

- the compiler
- the C# runtime (with and without intermediate JSON step)
- the JS runtime

The current smattering of tests in the C# project are designed to test the first two points to varying degrees and the suite is written using NUnit which is a C# specific thing.

However I can imagine an external test system, perhaps in node, which runs the C# binaries as well as inkjs...

joethephish:
Although there are problems with my suggestion:

- It might(?) not work with travis, which is doing continuous integration tests in the main repo, or would have to be rethought at the very least
- JS tests would initially fail if it was a shared system and we write new C# features.

erkyrath:
> JS tests would initially fail if it was a shared system and we write new C# features.

Well, that ought to be flagged, right? As long as it doesn't prevent the C# features from being checked in.

My experience is that you want a separate test system for the compiler. For that, you might want to test the contents of the generated JSON, whereas the runtime tests will be purely focused on the output of the runtimes.

(Of course, my experience is with the Inform ecosystem, whose testing is utterly haphazard and scattered all over the place.)

joethephish:
>> JS tests would initially fail if it was a shared system and we write new C# features.

> Well, that ought to be flagged, right?

Good point!

> My experience is that you want a separate test system for the compiler. For that, you might want to test the contents of the generated JSON, whereas the runtime tests will be purely focused on the output of the runtimes.

Hrm, yeah, good point, and I agree with you in theory, although I'm not too interested in the details of the JSON spec itself (it's the least important part of the pipeline), and I'm wary of losing the advantages I get from writing a simple ink-based test, and testing the result at the final output end of the pipe - the current test suite runs the test twice, one direct using in-memory runtime generated objects, and one making a round-trip via JSON. I've found it's quite a nice way to get good general coverage.

I'd rather not have to hand-author (or more likely, capture) JSON examples for everything I want to test in the runtime.

Conceptually it's quite a simple extension to convert the tests so they also test the JS runtime, although they'd all have to be converted to a new test environment.

Of course, this is all theoretical, because I doubt I'll have time to do such a thing any time soon :)

y-lohse:
I think a common ink test suite would be great, but I also don't have the time to research how to do such a set up, and I've waisted enough time thinking about test structures — I'm definitely at the "done is better than perfect" stage on this problem. So I'll go ahead with my initial plan, but count me in if something happens on this front.

I've just finished setting up the whole thing — there's one very basic test, but it is automatically run by travis, on both the default and es2015 flavors, and there is a coverage report (10% right now!).
Next step is to actually start writing the ink file that will contain all the tests. Progress will likely be just as slow as it has been so far though :(

y-lohse:
Ticked a few boxes, found a bug with glue (but not #79 so there's definitely something wrong there), and code coverage is up to 34% now 🕺 
This has become my official train-ride activity, so things will progress at the rate of my travels 🚋 ✈️

y-lohse:
@joethephish I might have found a rather harmless bug with variable observers, but not sure it's one, maybe you know this off the top of your head:

```
= variable_observer_stitch
VAR observedVar = 1
declared
~ observedVar = 3
mutated
```

I have an observer on `observedVar`. If I advance the story up to `declared`, my variable observer is called once, with a value of `1`. `1` is correct, but the value hasn't changed, so intuitively it shouldn't call the observer at all.

I assume that it's the same deal as with external functions: after reaching `declared`, the engine keeps going to see if the next output is glued, so it processes the rest of the stitch. The variable gets flagged has having changed, no glue is found so the engine steps back… and calls the observer function because `observedVar` is stil flagged as having changed, even though the value is the same? Or are there supposed to be safeguard against this and I missed them?

joethephish:
Hmm, nope this is definitely a bug. It works differently to external functions - the “diff” is only run for observed variables after the end of a Continue turn.

This particular case is a bit odd though. VARs are actually fully global and “static”, so your example should be the same as having the full VAR line including assignment at the top, outside the scope of the knot. (We just allow declarations anywhere for convenience.)

And I think variable observation is turned off during the initial globals setup, so I think you should actually get zero calls when it’s 1. But I’d have to check the code to remind myself (not at my computer right now).

joethephish:
Ah, no, I’m wrong - the observation does look like it’s run during initial globals setup. So I think that’s perhaps the call you’re seeing - when the VAR goes from its uninitialised value of 0 to 1. So yeah, arguably we could disable variable observation during that phase (the function is called ResetGlobals).

y-lohse:
Cool, thanks for the insight. I'm not quite sure it's the init call I'm seeing, but that's easy enough to check. Could quite possibly be a bug in inkjs.

y-lohse:
@joethephish puzzled by something new: when is [`InkList.AddItem`](https://github.com/inkle/ink/blob/e0db89f15e4ad7c8f99be81cd6d03db7a3660a6d/ink-engine-runtime/InkList.cs#L146) used? Looking at the code, I don't see it being called at all, and it's escaping my test coverage so far. It doesn't seem to be an override from `Dictionary` either..?

joethephish:
If it’s not called from anywhere it’s possible it’s just an API for runtime game code. Especially since it has a summary code doc and is marked as public rather than internal. (Can’t actually remember for sure off hand)

y-lohse:
Ok, thanks for the quick answer!

So, general status report: I've added tests for pretty much everything mentioned in the documentation, and the coverage is at 72%. A few tests are disabled because of bugs. Looking at the source, the main things that aren't covered yet:

- exceptions and error handlings
- the `Path` class
- casting variables
- container and choice flags
- `BuildStringOfHiearchy`
- math operations, comparisons and using && and || on some data types
- Turning things into JSON (diverts, glues, control commands, etc)
- `Random` and `SeedRandom`, `ListFromInt`

I'll add tests for some of these, others are very time consuming and I'll probably leave them out for now.

y-lohse:
Alright, time to wrap things up. Current coverage is 77%, when I fix the few bugs I found and re-enable the tests it should climb up to 80-ish. The 20 remaining percent require a lot of effort and not much mayoff, so they will have to wait.

PR #72, opened 2017-06-30
Correcting case of toJson to ToJson

russellquinn:
The case of the ToJson function is wrong. It's different from the C# implementation of Ink and from their documentation.


y-lohse:
I apologize for leaving this PR unattended for so long, it's totally legitimate.

PR #76, opened 2017-09-10
Saving of InkList is not working correctly

losttrout:
If you look at get fullName(), there's supposed to be a "." between the origin and the item names. InkLists can't be loaded correctly back from JSON without the period. 

There are two toString() methods in this class, but I'm assuming that was intentional.

### DIFF
--- a/engine/InkList.js
+++ b/engine/InkList.js
@@ -42,7 +42,7 @@ export class InkListItem{
 		if (this.originName != null)
 			originCode = this.originName.toString();
 		
-		return originCode + itemCode;
+		return originCode + "." + itemCode;
 	}
 }
 


y-lohse:
Ouch! Thanks, but this isn't the right fix — although I don't know what is, at this point. The double `toString` isn't intentional.
The one with the dot is the one that should be invoked when saving / loading from JSON, but I guess the second one (without dot) overrides it and causes the bug. Unfortunately the second one is used to convert a `InkListItem` into a key for the `InkList` hash.

I think the proper way to fix this is to use a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for `InkList` — Maps being javascript's version of C#'s dictionary. Browser support for Map seems ok, but it's a fair amount of code to change…

losttrout:
1. You could use fullName instead of toString() when decoding the JSON.

2. I think it is safe to use the '.' in the InkList hash too. My understanding is that originCode + '.' + itemCode in a globally unique string that identifies an InkListItem. originCode + itemCode actually isn't a globally unique string and might not be safe to use.

y-lohse:
You're right, although both would introduce discrepancies with the C# code which eventually becomes a pain in the ass.
But I probably won't have time to do the big rewriting soon, so I'm going to merge this PR and open an issue  so I remember the details later. Thanks for the fix!

Issue #80, opened 2017-11-22
ink 0.7.5 update


joethephish:
Not yet released but pending!

Full diff: https://github.com/inkle/ink/compare/0.7.4...0.7.5

y-lohse:
@joethephish Just a head's up, I noticed [`Path. componentCount`](https://github.com/inkle/ink/compare/0.7.4...0.7.5#diff-023e86641747b0b7c33d16bdfa816c30R80) and [`Path.length`](https://github.com/inkle/ink/compare/0.7.4...0.7.5#diff-023e86641747b0b7c33d16bdfa816c30R121) are redundant, not sure if that's intentional.

joethephish:
Not intentional, good catch! Guess I’ll remove componentCount, length is nicer :)

PR #81, opened 2017-11-26
Update to 0.7.5

y-lohse:
See #80

### DIFF
--- a/engine/CallStack.js
+++ b/engine/CallStack.js
@@ -4,6 +4,7 @@ import {Path} from './Path';
 import {StoryException} from './StoryException';
 import {JsonSerialisation} from './JsonSerialisation';
 import {ListValue} from './Value';
+import {StringBuilder} from './StringBuilder';
 
 class Element{
 	constructor(type, container, contentIndex, inExpressionEvaluation){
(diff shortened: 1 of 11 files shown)


y-lohse:
Left do do:

- [ ] callStackTrace  https://github.com/inkle/ink/compare/0.7.4...0.7.5#diff-188c261d8349543cc45d3f8a09c52143R358
- [ ] Profiler https://github.com/inkle/ink/compare/0.7.4...0.7.5#diff-d4e5d16fb4e24576d9d8454692aa957aR1
- Story profiler hooks : https://github.com/inkle/ink/compare/0.7.4...0.7.5#diff-39a71ebe9476df71b001ffd1ab0e8dfdR105

joethephish:
I wonder if there's some way we could categorise certain ink runtime features that are non-essential. The above features on your todo list aren't necessary for compatibility with ink 0.7.5 in general, there's definitely "would be nice" - it's possible that a JS-based profiler might not end up getting used very frequently at all.

Arguably this is my bad - I could've included the stubs (story profiler hooks), and then designed the API so that you have construct and pass in your own custom profiler that implements certain timing functions. The profiler itself could then have been added only to the ink-unity-integration package for example.

y-lohse:
I wouldn't mind that :) Things like `BuildStringsOfHiearchy` would fit in there nicely too.

I've decided to postpone implementing the profiling code — as you say it's not essential, and porting it is a bit tricky and eats up time I'd rather spend fixing bugs and updating the tooling. 

Have you already "officially" release v0.7.5 of ink? If so I'll also publish an update for inkjs.

PR #101, opened 2018-01-23
Unset the "in external evaluation" flag

joningold:
On completion of an external function, unset the flag that tells the engine we're in an externally called function.

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -681,6 +681,9 @@ export class StoryState{
 		this._originalEvaluationStackHeight = 0;
 		
 		this._variablesState.callStack = this.callStack;
+		
+		// No longer in external function eval
+	        this._isExternalFunctionEvaluation = false; 
 
 		if (returnedObj) {
 			if (returnedObj instanceof Void)


y-lohse:
Thanks! A bit worrying that we didn't have this line but already had [that one](https://github.com/inkle/ink/commit/692b3641352e751a3a34eb302a2da6a2ee1ff8dc#diff-13b72c7a8367280798bdbe866c3d5ef2R267).

joningold:
Yes! Somehow the first change was found in ink-js months ago, but never
successfully ported into the main repo - then we found and fixed it again,
more properly, yesterday, and I was surprised to find half the fix in js!
ARGH!

joethephish:
And just to make things more complicated - I've now pushed a commit to the ink repo which rewrites all this ;)

https://github.com/inkle/ink/commit/3afd463bc96779429b6463868a62a8854d638da3

(I'm much much happier with the new approach, it's more stable, flexible and less hacky.)

y-lohse:
Meh :D

I'd be very interested in adding test coverage for that, especially if it's going to be rewritten. So if you have a program that reproduces a bug, I'd be super interested!

joethephish:
Hrm, I'm afraid that particular bug was found by stepping deep through our Heaven's Vault ink, sorry!

joethephish:
I just added a test to the C# suite... here it is in all its confusing glory!

The ink:

```
EXTERNAL gameInc(x)

== function topExternal(x)
In top external
~ return gameInc(x)

== function inkInc(x)
~ return x + 1
```

And the C#:

```
// Crazy game/ink callstack:
// - Game calls "topExternal(5)" (Game -> ink)
// - topExternal calls gameInc(5) (ink -> Game)
// - gameInk increments to 6
// - gameInk calls inkInc(6) (Game -> ink)
// - inkInc just increments to 7 (ink)
// And the whole thing unwinds again back to game.

story.BindExternalFunction("gameInc", (int x) => {
    x++;
    x = (int) story.EvaluateFunction ("inkInc", x);
    return x;
});

string strResult;
var finalResult = (int) story.EvaluateFunction ("topExternal", out strResult, 5);

Assert.AreEqual (7, finalResult);
Assert.AreEqual ("In top external\n", strResult);
```

y-lohse:
Riiiiight, of course 😅 . Thanks, I'll integrate that in the ink.js test suite too !

PR #115, opened 2018-02-22
Add test for external function sandwich

y-lohse:
Trying to add the test for [this infamous issue](https://github.com/y-lohse/inkjs/pull/101#issuecomment-360435744) (#102), but haven't got it working yet. Seems like there's another bug somewhere?

1. We call `story.EvaluateFunction('topExternal', [5], true);`
1. The story `Continue`s and finds a new line after `In top external`, so it creates a copy of the stack
1. The external function `gameInc` is called
1. `gameInc` calls `story.EvaluateFunction('inkInc', [x])`
1. When it's done evaluating `inkInc`, the `state._isExternalFunctionEvaluation` is set to `false`
1. It goes to `state.TryExitExternalFunctionEvaluation`, which returns `false` because `_isExternalFunctionEvaluation` is `false`
1. The whole thing crashes because [the stack can't be popped](https://github.com/y-lohse/inkjs/blob/master/engine/Story.js#L679)

I'm not sure what's wrong where yet, but any insight is welcome!

I should check when exactly `_isExternalFunctionEvaluation` is set to false, and whether we're really done with `gameInk` or not. And if we really are, maybe there's a snapshot that should be restored?


joethephish:
This definitely can't work until you integrate latest from the C# repo - [this commit](https://github.com/inkle/ink/commit/3afd463bc96779429b6463868a62a8854d638da3) rewrites the way that external function evaluation is done so that the "sandwich" works correctly (and there are also additional fix commits that come after it).

e.g. `_isExternalFunctionEvaluation` doesn't exist anymore!

y-lohse:
Oh ok, I didn't get that last time. Thanks!

Issue #126, opened 2018-04-20
[Question]: Are you planning to upgrade to new 0.8.0 version of Ink?


fireton:
Just wondering. I'm working on a small ink game and choosing a platform to implement it.

Tom-Julux:
Most of the improvements in Version 0.8 are on the compiler side. You can use the current inkjs version with the ink 0.8 compiler in most cases. However you have to change the version of your json file manualy.

from
```json
{"inkVersion":18}
```
to 
```json
{"inkVersion":17}
```

Then everything should work as expected

erkyrath:
In that case, it's worth updating inkVersionCurrent to 18.

joethephish:
While @Tom-Julux's little hack may work in many cases, there's a good reason I bumped the version number - the story format has changed slightly. So while it might work for your story, I'd say you should only do so at your own risk! At a guess though, the main errors you're likely to see are to do with newline handling, which may not be the end of the world for most people.

joningold:
We recommend you use inky; the current build will let you work on latest ink for Unity, with the export to web option falling back to the older version of ink-js. There are few unsupported features / changes, but at least it's less confusing.

y-lohse:
Hi all! Sorry for the delay, I was offline for a while. But now that I'm back, yes, I certainly plan on updating inkjs. Help is of course always appreciated, but I haven't looked into what needs to be done yet. I'll create a work in progress PR soon-ish.

PR #129, opened 2018-04-26
Update engine/Story.js to inkVersion v18

BTBTravis:
Some folks were requesting v18 inkVersion support on issue #126  so I went through and searched for 17's.
> `grep -r -n --include=\*.js "17" .`
```
./templates/browser_serverless/ink.js:1929:  	Done: 17,
./templates/browser_serverless/ink.js:4256:  //Taken from https://gist.github.com/blixt/f17b47c62508be59987b
./templates/browser_serverless/ink.js:5134:  		_this.inkVersionCurrent = 17;
./templates/browser_with_server/ink.js:1929:  	Done: 17,
./templates/browser_with_server/ink.js:4256:  //Taken from https://gist.github.com/blixt/f17b47c62508be59987b
./templates/browser_with_server/ink.js:5134:  		_this.inkVersionCurrent = 17;
./engine/Story.js:34:		this.inkVersionCurrent = 17;
./engine/PRNG.js:1://Taken from https://gist.github.com/blixt/f17b47c62508be59987b
./engine/ControlCommand.js:104:	Done: 17,

```
`./engine/Story.js` seemed like the only file that needed updating so I incremented that 17 to 18. 

Unfortunately when I saved the .js file with vim it changed the white space somehow so pardon all the needless changes. I'm still figuring out how to save without it changing those.

### DIFF
--- a/engine/Story.js
+++ b/engine/Story.js
@@ -28,20 +28,20 @@ if (!Number.isInteger) {
 export class Story extends InkObject{
 	constructor(jsonString, lists){
 		super();
-		
+
 		lists = lists || null;
-		
-		this.inkVersionCurrent = 17;
+
+		this.inkVersionCurrent = 18;
 		this.inkVersionMinimumCompatible = 16;
-		
+
 		this._variableObservers = null;
 		this._externals = {};
 		this._prevContainerSet = null;
 		this._listDefinitions = null;
-		
+
 		if (jsonString instanceof Container){
 			this._mainContentContainer = jsonString;
-			
+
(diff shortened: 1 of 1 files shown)


y-lohse:
Thank you! `./engine/Story.js` is indeed the one and only place where it should be changed manually.


But as you may have seen on #126 , [we can't just bump that number and leave the rest unchanged](https://github.com/y-lohse/inkjs/issues/126#issuecomment-383483878), or people will run into errors. This will be part of a larger PR at some point though.

PR #132, opened 2018-05-03
Added string !? substring

joningold:


y-lohse:
Thanks! Is this part of the v0.8 changes? If so I'll leave it unmerged until the larger work on that has started.

joningold:
It’s one of the little fixes we added to 0.8.0 without bumping the version
number so, um, yes. Take it whenever suits!

Jon

Issue #133, opened 2018-05-07
v0.8.1 upgrades


y-lohse:
Following up on #126, here's the diff : https://github.com/inkle/ink/compare/0.7.5...0.8.1#diff-188c261d8349543cc45d3f8a09c52143

PR #139, opened 2018-05-17
Ink v8

y-lohse:
Tackling #133.

Here's the list of changes, in order of appearance.
Small changes are 🐞 , large ones are 🐘 

- [x] Callstack Pointers
- [x] Choice and ChoicePoint 🐞 
- [x] Container Searchresults 🐞
- [x] Divert pointers
- [x] Glue 🐞
- [x] Serilization changes 🐞
- [x] Object 🐞
- [x] Path PathByAppendingComponent 🐞
- [x] Pointer
- [x] PushPop 🐞
- [x] Searchresult 🐞
- [x] Story (pointers, choices, …) 🐘 
- [x] StoryState (pointers, glue, …) 🐘 
- [x] Value BadCastException 🐞
- [x] VariableReference 🐞
- [x] VariableState
- [x] Fix failing tests
- [x] Check test coverage (previously ~80%)

The changes are pretty significative, so any help is appreciated! Just let me know here what you'd like to work on so we can coordinate.


NQNStudios:
I'm interested in helping! I'll look at some of the changed portions later today and see if any look doable for starters. Will I need to make a separate PR of my own?

y-lohse:
Awesome! The list above mirrors the file list in the diff, and anything marked with a  🐞should be do-able. But if you need help with anything, please let me know, I'd be more than happy to help.

For the workflow, I suggest you create a new branch based on `v8`, and then create a PR targeting the `v8` branch again. That way we can keep track of all the changes in this "main" PR here, while merging things as we go.

ephread:
Thanks a lot for maintaining the port, I'd like to help too! What's the easiest 🐞 I can tackle to get started? I'll take a look at the diff to warm me up.

I have a couple of additional questions:

- is this a straight human-transpiler task or is there some fancy stuff to know?
- should we make sure that tests keep passing at every step, or complete all the tasks and then rework the tests?
- is there any coding style we should be following?

joethephish:
Out of interest, do you usually try to port one commit at a time, or do you port the whole diff in one go?

I can see pros and cons - e.g. if I try something and revert it a few commits later if's a nightmare if you're going one commit at a time...! On the other hand, it's probably easier to keep track of changes in most other cases, especially if code is moved around.

I've been half wondering about doing a bit of code shuffle - splitting up files for better organisation and code clarity (e.g. Story.cs is huge). But I'm slightly resisting changes like these in case it makes it a nightmare to port to inkjs!

y-lohse:
@ephread wow, awesome, thank you! Let's see:

> is this a straight human-transpiler task or is there some fancy stuff to know?

It's pretty much human-transpiling, but as you've figured out we try to stick as closely as possible to the C# code. It makes maintenance in the long run a lot easier :)

> should we make sure that tests keep passing at every step, or complete all the tasks and then rework the tests?

In this case, since there's a large amount of changes, I planned to fix them all at once at the end. Mostly because the conversion doesn't happen in a logical order.

> is there any coding style we should be following?

I saw you tried your best at figuring that out, so thanks 😛 . I don't trust unspoken code-style rules, so feel free to do as you see fit. I'll set up an automatic linter/formatter so that no one has to care too much about this in the future.

The diff looks manageable! I'll create the PR and we can keep working there.

y-lohse:
@joethephish I port the diff in one go, I've never tried to do it commit by commit — although it makes sense for smaller change-sets, I might try that.

I'd be quite in favor of a code shuffle actually :-). As long as there are no other changes in the same commit, it'd be easy enough to port, and it would make things easier in the long run — finding where things need to be changed in the larger files is a bit tedious.

ephread:
> I'll set up an automatic linter/formatter so that no one has to care too much about this in the future.
Sounds great, let's let the computer do the heavy lifting. 🤖

> The diff looks manageable! I'll create the PR and we can keep working there.
Awesome!

ephread:
@y-lohse I didn't realize your commit messages were following the semantic-release style. It's duly noted, I'll make my future commits behave. 😉

ephread:
Now that all the PR are en-route, I created [a branch](https://github.com/ephread/inkjs/tree/v8-all-the-pr) in which I merged all the changes and I'm trying to make all the tests pass.

18 tests fail at the moment and I believe they all fail for the same reason. I'm not familiar enough with the internals to express the issue I've noticed in an idiomatic manner, but it seems that some instructions get eaten up. That results in a execution cut short, or missing new lines.

I haven't been able to track down the issue, but it's most likely some very, very stupid typo/oversight, well hidden somewhere. If anyone wants to weight in or has any idea where the issue might lie, that would be a huge help!

y-lohse:
I'll take a look as soon as the various PRs are merged!

Starting a posible todo list for after v8:

- switch all the `var as Thing` statements to the ternary form, or possibly introducing a `castAs(myVar, Class)` helper function
- `path.length` should be `path.Length` to be consistent with C#

ephread:
Great!

I've discovered a number of other minor typos that I've introduced and I'll make another PR for these (specifically, in Callstack).

I'm down to 14 failing tests, but I haven't had the chance to update the branch. I'm off to bed, good luck!

y-lohse:
Alright, I'll wait for that PR. I also see that something is going wrong in Callstack but since you've already identified the issue, I'll spare myself the digging.

Some tests are failing because ink used to return new lines at the end of some outputs, and doesn't seem to do it anymore. So it's possible we just have to update some test results to fix those.

ephread:
> Some tests are failing because ink used to return new lines at the end of some outputs, and doesn't seem to do it anymore. So it's possible we just have to update some test results to fix those.

I'm not sure that the missing new lines are new expected behavior. For instance, If you take a look at `logic.casts`, `{intvar + 0.5}` returns `521.5\n` but `{floatvar:float var is truthy|is falsy and broken}` returns `float var is truthy`. Its seems that, when some form conditional logic is involved, the new lines get eaten up.

y-lohse:
Ok I've spent a lot of time investigating some of the failing tests… and it turns out everything is operating as expected. Looping in @joethephish for confirmation.

[Here's the offending ink](https://github.com/y-lohse/inkjs/blob/master/tests/inkfiles/tests.ink#L40-L48):

```
= if_text_truthy
~ temp met_blofeld = true
I… {met_blofeld: I saw him. Only for a moment.}
-> DONE
```

The output in ink v7 was  [`I… I saw him. Only for a moment.\n`](https://github.com/y-lohse/inkjs/blob/master/tests/specs/ContentSpec.js#L34). Now, it's `I…  I saw him. Only for a moment.`

First up, cleaning white spaces — i believe there is a regression with ink v8. There is now a double white space after `I…`. `CleanOutputWhiteSpace` copies all the output from the first whitespace [it encounters](https://github.com/inkle/ink/blob/master/ink-engine-runtime/StoryState.cs#L191), so they both get copied.

Then, more tricky, the missing `\n`. I'm still not sure what the behavior of ink C# v8 is on this. The json looks like this:

```
"if_text_truthy": ["ev", 1, "/ev", {
        "temp=": "met_blofeld"
      }, "^I… ", "ev", {
        "VAR?": "met_blofeld"
      }, "/ev", [{
        "->": ".^.b",
        "c": true
      }, {
        "b": ["^ I saw him. Only for a moment.", {
          "->": ".^.^.^.9"
        }, null]
      }], "nop", "\n", "done", {
        "#f": 3
      }],
```

Since the condition is truthy, we get diverted to `^.^.^.9` which resolves to an in-existing section of the parent container, the pointer resolves to `null` and everything stops. The `"nop", "\n", "done"` doens't get evaluated as far as I can tell.

On the other hand if the condition is falsy, the divert doesn't happen and the `\n` gets added to the output stream. Is that the behavior in C#? I think `inklecate` eats up the new lines so it's hard to tell.

y-lohse:
For my own reference, other failing tests are:

- `Integration should not increment the read count as long as the knot is not left` -> read counts now gets incremented even when the knot isn't left, need to check if that's intended
- `Content should print handle an if/else text`simply fails
- `should find line by line tags` reads too many tags (reads the ones from the next line)
- `Simple lists should handle blank elements` fails outright

y-lohse:
Ok, we're done here, time to push that button!

Huge thanks again @ephread , all your contributions are very appreciated. We definitely wouldn't be here without you!

ephread:
@y-lohse no worries! Thank you for maintaining the project!

> Wooooof finally found the origin of the new line problem, all the old tests are working again \o/.

Out of curiosity, what was the culprit?

y-lohse:
[It was this bit](https://github.com/y-lohse/inkjs/pull/139/commits/fa73a941b952d2f6dec50ea35e25ea16ef5d70f8). Once an ink file is compiled to JSON, it becomes a tree structure with a lot of the outermost leaves being diverts to some other content. When the divert path is converted to a pointer, this code runs — but often times the target is not a `Container` but maybe a `ControlCommand` or whatever. So in these cases we would stop the loop early and return the wrong content, eventually end up with a `null` pointer and stop the execution too early.

ephread:
Nasty! It must have been so hard to debug; again, thanks a lot for uncovering and fixing the issue, it really did drive me crazy for a while 🤯.

joethephish:
Sorry I'm late to the party on this one, but just saw your solution, that's great! On the good side, this may be the kind of thing that TypeScript might help with in future :)

PR #143, opened 2018-05-25
Fix for ListFromInt

joningold:
The intVal parameter was already casted to a normal int, but then trying to use ".value" to read its value.


y-lohse:
Ah yes, good catch, thanks!

PR #146, opened 2018-06-05
Fix #145

### LINKED ISSUE #145: Shuffle Lists
Hi,
was playing around with procedural text generation in ink (like https://www.patreon.com/posts/tips-and-tricks-18636280 or https://gist.github.com/joethephish/2b0a291806b3100390fdddd56defd169)
which is a lot of fun, but got a weird problem with inkjs and shuffle lists.

```
-> conversation_loop

=== conversation_loop 
	+	[Test Shuffle] -> test_shuffle
	+	[Done] -> test_done

=== test_shuffle
    {first()} and {second()}
    -> conversation_loop
    
===  test_done
    Bye :)
    -> END

== function first ==
{~
 - A
 - B
 - C
 - D
 - E
 }
 
 == function second ==
{~
 - A
 - B
 - C
 - D
 - E
 }
```
When playing this ink in inky or inklecrate you get the expected
A and E
B and C
...

Playing with inkjs you get
E and E
B and B
A and A
...

I traced this down to `NextSequenceShuffleIndex()`
For both shuffle lists this function gets called with exactly the same state/values, and so returns the same choice. This problem only occurs with lists of the same length, because the length is part of the random seed calculation (so don't use lists of the same length in the same place and you are fine).

I do not understand, why the c# implementation behaves differently. The code nearly looks the same. But I can't debug the c# code.

My problem could be fixed with something like

```
NextSequenceShuffleIndex(){

	var numElementsIntVal = this.state.PopEvaluationStack();
	if (!(numElementsIntVal instanceof IntValue)) {
		this.Error("expected number of elements in sequence for shuffle index");
		return 0;
	}

	var numElements = numElementsIntVal.value;
			
	let min = 0
	let max = Math.floor(numElements);
	// exclusive MAX
	let chosen=  Math.floor(Math.random() * (max - min)) + min;

	return chosen
}
```

which works fine for me, but because you are losing the seedable random number and the 'sequence' functionality, this isn't really a patch.

NQNStudios:
I figured this might be a good first issue to tackle. After some testing, I found that the problem was a silent error in hashing the path of the sequence being shuffled. All hashes were evaluating to 0 because `charCodeAt[i]` attempts to array-index the function, `charCodeAt()`, returning undefined and making these line fail to sum anything:

```
for (var i = 0, l = seqPathStr.length; i < l; i++) {
    sequenceHash += seqPathStr.charCodeAt[i] || 0;
}
```

Which is why the same length sequence always ended up producing the same seed.

### DIFF
--- a/engine/Story.js
+++ b/engine/Story.js
@@ -1494,7 +1494,7 @@ export class Story extends InkObject{
 		var seqPathStr = seqContainer.path.toString();
 		var sequenceHash = 0;
 		for (var i = 0, l = seqPathStr.length; i < l; i++){
-			sequenceHash += seqPathStr.charCodeAt[i] || 0;
+			sequenceHash += seqPathStr.charCodeAt(i) || 0;
 		}
 		var randomSeed = sequenceHash + loopIndex + this.state.storySeed;
 		var random = new PRNG(parseInt(randomSeed));
--- a/templates/browser_serverless/ink.js
+++ b/templates/browser_serverless/ink.js
@@ -1,7 +1,7 @@
 (function (global, factory) {
   typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
-  typeof define === 'function' && define.amd ? define('inkjs', ['exports'], factory) :
-  (factory((global.inkjs = global.inkjs || {})));
+  typeof define === 'function' && define.amd ? define(['exports'], factory) :
+  (factory((global.inkjs = {})));
 }(this, (function (exports) { 'use strict';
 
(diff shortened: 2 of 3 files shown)


y-lohse:
That's fantastic, thank you very much! Absolutely looks like that would be the problem.

Only thing before we merge, can you please discard the changes made to the ink.js template files? I'll update them once we publish a new version :)

NQNStudios:
Done. Yeah, I figured I had done something wrong when my PR had >200 line changes after I only changed 2 characters, haha!

y-lohse:
Yeah no worries haha. Thanks again!

PR #160, opened 2018-07-02
Integrate @ephread changes

y-lohse:
See https://github.com/y-lohse/inkjs/pull/139#issuecomment-401754921

Thanks again for this @ephread ! I'll do a proper review as soon as possible — probably tonight.

### DIFF
--- a/engine/Container.js
+++ b/engine/Container.js
@@ -2,25 +2,26 @@ import {StringValue} from './Value';
 import {StoryException} from './StoryException';
 import {StringBuilder} from './StringBuilder';
 import {Object as InkObject} from './Object';
+import {SearchResult} from './SearchResult';
 
 export class Container extends InkObject{//also implements INamedContent. Not sure how to do it cleanly in JS.
 	constructor(){
 		super();
 		this.name = '';
-		
+
 		this._content = [];
 		this.namedContent = {};
-		
+
 		this.visitsShouldBeCounted = false;
 		this.turnIndexShouldBeCounted = false;
 		this.countingAtStartOnly = false;
-		
+
 		this.CountFlags = {
 			Visits: 1,
 			Turns: 2,
 			CountStartOnly: 4
 		};
-		
+
 		this._pathToFirstLeafContent = null;
 	}
 	get hasValidName(){
@@ -34,7 +35,7 @@ export class Container extends InkObject{//also implements INamedContent. Not su
 	}
 	get namedOnlyContent(){
 		var namedOnlyContentDict = {};
-		
+
 		for (var key in this.namedContent){
 			namedOnlyContentDict[key] = this.namedContent[key];
 		}
@@ -100,7 +101,7 @@ export class Container extends InkObject{//also implements INamedContent. Not su
 		return this._pathToFirstLeafContent;
 	}
 	get internalPathToFirstLeafContent(){
-    var components = [];
+		var components = [];
 		var container = this;
 		while (container instanceof Container) {
 			if (container.content.length > 0) {
@@ -111,7 +112,7 @@ export class Container extends InkObject{//also implements INamedContent. Not su
 		}
 		return new Path(components);
 	}
-	
+
 	AddContent(contentObj){
 		if (contentObj instanceof Array){
 			contentObj.forEach(c => {
@@ -120,11 +121,11 @@ export class Container extends InkObject{//also implements INamedContent. Not su
 		}
 		else{
 			this._content.push(contentObj);
-			
+
 			if (contentObj.parent) {
-                throw "content is already in " + contentObj.parent;
-            }
-			
+				throw "content is already in " + contentObj.parent;
+			}
+
 			contentObj.parent = this;
 
 			this.TryAddNamedContent(contentObj);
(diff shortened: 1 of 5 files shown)


ephread:
You're welcome! Great, let me know if you need me to make any change to the PR.

engine/Container.js
```diff
@@ -205,7 +221,7 @@ export class Container extends InkObject{//also implements INamedContent. Not su
 				return foundContent;
 			}
 			else {
-				throw new StoryException("Content '"+component.name+"' not found at path: '"+this.path+"'");
+				throw null;
```
y-lohse: [I think we just want to return `null` here](https://github.com/inkle/ink/compare/0.7.5...0.8.1#diff-ebd46ea8e5c494f6b2a75971f9ad2badR224), not throw it :)

engine/Divert.js
```diff
@@ -27,24 +28,31 @@ export class Divert extends InkObject{
...
 	}
-	get targetContent(){
-		if (this._targetContent == null) {
-			this._targetContent = this.ResolvePath(this._targetPath);
+	get targetPointer(){
+		if (this._targetPointer.isNull) {
+			var targetObj = this.ResolvePath(this._targetPath).obj;
+
+			if (this._targetPath.lastComponent.isIndex) {
+				_targetPointer.container = targetObj.parent;
```
y-lohse: On this and the next couple of lines, I think we want `this._targetPointer` instead of just `_targetPointer`.

engine/Pointer.js
```diff
@@ -0,0 +1,43 @@
+import {Path} from 'path.js';
+
+export class Pointer{
+	constructor(container, index){
+		this.container = container;
+		this.index = index;
+	}
+
+	Resolve(){
+		if (index < 0) return this.container;
```
y-lohse: We probably want `this.index` here :)

edit: same thing on the next couple of lines

engine/Pointer.js
```diff
@@ -0,0 +1,43 @@
...
+
+		return this.container.content[index];
+	}
+
+	get isNull(){
+		return this.container == null;
+	}
+
+	get path(){
+		if (this.isNull()) return null;
```
y-lohse: Since is null is a `get`, I believe we want `if (this.isNull)`, ie. not calling the function.

engine/Pointer.js
```diff
@@ -0,0 +1,43 @@
...
+
+		if (index >= 0)
+			return this.container.path.PathByAppendingComponent(new Path.Component(index));
```
y-lohse: We're missing the `this.index` here.

engine/Pointer.js
```diff
@@ -0,0 +1,43 @@
...
+		else
+			return this.container.path;
+	}
+
+	toString(){
+		if (!this.container)
+			return "Ink Pointer (null)";
+
+		return "Ink Pointer -> " + container.path.toString() + " -- index " + index;
```
y-lohse: We're also missing some `this` here ;)

engine/Pointer.js
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	StartOf(container){
```
y-lohse: `StartOf` seems to be a [static function](https://github.com/inkle/ink/compare/0.7.5...0.8.1#diff-732a76cbfef9b62d2ed80adc70882a62R59), so we need to declare it as `static StartOf` here as well.

engine/SearchResult.js
```diff
@@ -0,0 +1,20 @@
+// When looking up content within the story (e.g. in Container.ContentAtPath),
```
y-lohse: Would you mind removing these comments? I know I've done it myself, but in hindsight leaving comments from the original implementation is not great. To understand the code, reading the C# versions is preferable, and it drowns out comments that are specific to the js port.

engine/SearchResult.js
```diff
@@ -0,0 +1,20 @@
...
+// in the path to find the closest valid container. Instead of crashing horribly,
+// we might see some slight oddness in the content, but hopefully it recovers!
+export class SearchResult{
+	constructor(){
+		this.obj;
+		this.approximate;
+	}
+
+	get correctObject(){
+		return this.approximate ? null : obj;
```
y-lohse: We want `this.obj` here and in `get container` :)

engine/SearchResult.js
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	get container(){
+		return obj;
```
y-lohse: So in the [original code](https://github.com/inkle/ink/compare/0.7.5...0.8.1#diff-2c8172fa5b22cfd139d6362cdb03de52R16), `obj` gets casted to a `Container`, which makes it `null` or `falsy` if the cast fails (can't remember which one, but I think it's `null`.)

If we return `obj` as-is, I think we risk returning a non-null value to a calling code would expect `null` under some conditions. So maybe `return this.obj instanceof Container ? this.obj : null` would be preferable here. What do you think?

engine/VariablesState.js
```diff
@@ -254,18 +258,26 @@ export class VariablesState{
 	//the original code uses a magic getter and setter for global variables, allowing things like variableState['varname]. This is not quite possible in js without a Proxy, so it is replaced with this $ function.
 	$(variableName, value){
 		if (typeof value === 'undefined'){
+			// Search main dictionary first.
```
y-lohse: Ditto for comments here if you don't mind.

ephread:
Thanks for taking the time to review and for catching these issues, I'm super sorry about all the missing `this`. I primarily code using Swift and Typescript, so I guess I lost the habit.

I'll fix all the mistakes in the coming hours!

engine/Container.js
ephread: Totally, that's a typo! :)

engine/Pointer.js
ephread: Typo again! I wish VS Code would warn me about that stuff, I guess I didn't configure it properly.

engine/Pointer.js
ephread: Good catch, that's an oversight.

engine/SearchResult.js
ephread: Yup, no problem.

You know what, I'll make another PR to add CONTRIBUTING.md detailing how to deal with this sort of thing. Comments, porting-style, etc.

engine/SearchResult.js
ephread: You're totally right. I totally missed the difference between the two types of cast (`(Container)` and `as Container`). I've checked and indeed, `null` would be returned if `obj` wasn't a `Container`. I'll fix it.

ephread:
I've ported (hopefully, I didn't miss as many `this` this time) the Callstack. It's in its [own branch](https://github.com/ephread/inkjs/tree/v8-callstack). Do you want me to include it in this PR?

That was, I believe, the last remaining file before tackling the big elephants (Story & StoryState).

ephread:
I just realized that we're losing the value-semantic of `struct Pointer`. While they will be copied when passed around in the C# code, in Javascript, it's the same objects that will be passed around, resulting in some nasty side effects.

I'm going to check all `Pointer` assignments and make sure that we're copying the pointer when necessary (which will introduce an extra copy method on `Pointer`). If you have a better option, let me know!

ephread:
No worries. 😉

> Is there anything else you wanted to do as part of this PR?

Nope, if all issues have been fixed, let's merge! 🎉

> For the `Callstack`, I think a new PR would be easier in terms of reviewing.

Got it, I'll open the PR once I'm out of `StoryState.js` (which is a significant PITA, I'm in no hurry to look at `Story.js` 😂 and I will definitely need an extra set of eyes to catch all the issues).

y-lohse:
Excellent, thanks a lot for tackling all this! 

If I can manage I'll try to get the tests running again so we get a third set of eyes :)

PR #164, opened 2018-07-03
fix: Path class import path

y-lohse:
To be merged after #162 to avoid rebasing.


y-lohse:
This has been taken care off elsewhere.

Issue #161, opened 2018-07-02
Enforce code styles


y-lohse:
A consistent coding style should be automatically enforced on this repo. Rough list of things to do:

- add an `.editorconfig` file
- add eslint and configure it so that we can keep a code-style that's close to the original C#
- add a `lint` task and have travis run it

ephread:
By the way, do you use any specific style guide?

Let's take advantage of this issue to lay out all the rules that you'd like us to follow and ideally, if it's a well-known style, all the formatting rules will have already been laid out in ready to use packages.

Also, I don't believe that we should necessarily follow the formatting of the C# code. The structure has become sufficiently different so that line numbers and function order don't match. And, as long as the signatures match, it's fairly easy to recognize the similar patterns in both Javascript and C#.

We should definitely pick a style and make all the code conform to it.

y-lohse:
So normally, I'd be in favour of just using an existing style guide and not spending much time setting up specific rules and stuff.

But while I hear your argument on the C# structure, I tend to disagree. Line numbers aren't too important, but I've found that keeping the same order and structure as much as possible is immensely helpful in the long run. When looking at a C# diff, it's a lot easier when you can just scroll along in the js/ts code, instead of searching for function names and jumping around the code.

What I had in mind was essentially taking `eslint:recommended` as a starting point, and then decide on a case by case basis for the rules that do break if we should change the rule or the code. i got s far as this:

```
"indent": [
            "error",
            2
        ],
        "linebreak-style": [
            "error",
            "unix"
        ],
        "quotes": [
            "error",
            "single"
        ],
        "semi": [
            "error",
            "always"
        ],
        "no-cond-assign": "off",
        "no-constant-condition": ["error", { "checkLoops": false }],
        "no-console": "off"
```

So typically `no-cond-assign` and `no-constant-condition` allow us to stay closer to the C# code, so I changed the rules. But errors about `var`s being re-assigned are a good point and don't actually impact legibility, so the code should be fixed on those.

As I said in the other issue, there are some rules that I actually feel somewhat strongly about — ie. not abusing `const` or keeping line numbers (it makes me loose the habit of using them and then I wonder why my code in other languages crash). On the other hand, I don't care at all about indentation styles or linebreaks.

I hope this gives at least a general idea of what I had in mind? I think that as part of #175 , we should keep linting related changes to a minimum, and we can iterate on that.

y-lohse:
[...]

PR #165, opened 2018-07-03
Code style

y-lohse:
Addresses #161 , to be continued after #139 .

Once everything is auto-fixed there are still 100 or so errors to go through. We'll have to see if we want to keep the console statements around or not.


y-lohse:
This will be taken care of by #175

PR #168, opened 2018-07-04
feat: update StoryState with v8 changes

ephread:
Thanks again for reviewing, we're getting there!

I have added a couple of comment, to highlight certain changes.

### DIFF
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -650,7 +650,7 @@ export class StoryState{
 
 		this._currentTurnIndex++;
 	}
-	StartExternalFunctionEvaluationFromGame(funcContainer, args){
+	StartFunctionEvaluationFromGame(funcContainer, args){
 		this.callStack.Push(PushPopType.FunctionEvaluationFromGame, this.evaluationStack.length);
 		this.callStack.currentElement.currentPointer = Pointer.StartOf(funcContainer);
 
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -369,11 +369,11 @@ export class StoryState{
 		}
 		else{
 			if(numberOfObjects > this.evaluationStack.length) {
-			throw "trying to pop too many objects";
-		}
+				throw "trying to pop too many objects";
+			}
 
-		var popped = this.evaluationStack.splice(this.evaluationStack.length - numberOfObjects, numberOfObjects);
-		return popped;
+			var popped = this.evaluationStack.splice(this.evaluationStack.length - numberOfObjects, numberOfObjects);
+			return popped;
 		}
 	}
 	PeekEvaluationStack(){
--- a/engine/StoryState.js
+++ b/engine/StoryState.js
@@ -12,12 +12,13 @@ import {JsonSerialisation} from './JsonSerialisation';
 import {Story} from './Story';
 import {PRNG} from './PRNG';
 import {Void} from './Void';
+import {Pointer} from './Pointer';
 
 export class StoryState{
-	constructor(story){		
+	constructor(story){
 		//actual constructor
 		this.story = story;
-		
+
 		this._outputStream = [];
 		this._outputStreamTextDirty = true;
 		this._outputStreamTagsDirty = true;
@@ -31,8 +32,8 @@ export class StoryState{
 		this._visitCounts = {};
 		this._turnIndices = {};
 		this._currentTurnIndex = -1;
-		
-		this.divertedTargetObject = null;
+
+		this.divertedPointer = Pointer.Null;
 
 		var timeSeed = (new Date()).getTime();
 		this.storySeed = (new PRNG(timeSeed)).next() % 100;
@@ -42,12 +43,9 @@ export class StoryState{
 		this._currentText = null;
 		this._currentTags = null;
 		this._currentErrors = null;
-		
+		this._currentWarnings = null;
+
 		this.didSafeExit = false;
-		
-		this._isExternalFunctionEvaluation = false;
-		this._originalCallstack = null;
-		this._originalEvaluationStackHeight = 0;
 
 		this.GoToStart();
 	}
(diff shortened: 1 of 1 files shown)


engine/StoryState.js
```diff
@@ -312,31 +299,45 @@ export class StoryState{
...
+			var c = st.charAt(i);
+
+			var isInlineWhitespace = (c == ' ') || (c == '\t');
+
+			if (isInlineWhitespace && currentWhitespaceStart == -1)
+				currentWhitespaceStart = i;
+
+			if (!isInlineWhitespace) {
+				if (c != '\n' && currentWhitespaceStart > 0) {
+					sb.Append(str.substr(currentWhitespaceStart, i - currentWhitespaceStart));
```
ephread: This certainly needs a more thorough check, `substr` should do exactly the same thing as `Substring`, but we can't be too careful.

engine/StoryState.js
```diff
@@ -385,15 +386,20 @@ export class StoryState{
...
+				this.OutputStreamDirty();
 				return;
 			}
 		}
 
 		this.PushToOutputStreamIndividual(obj);
 		this.OutputStreamDirty();
 	}
+	PopFromOutputStream(count){
+		this.outputStream.splice(this.outputStream.length - count, count);
```
ephread: Same here with `splice`, it should do exactly the same as `RemoveRange` but a double check would not be bad.

engine/StoryState.js
```diff
@@ -518,31 +549,16 @@ export class StoryState{
...
-//			var glue = obj as Glue;
-			var glue = obj;
-
-			if (cmd instanceof ControlCommand || (txt instanceof StringValue && txt.isNonWhitespace)) {
-				foundNonWhitespace = true;
-				if( rightGlueToStopAt == null )
-					break;
-			} else if (rightGlueToStopAt && glue instanceof Glue && glue == rightGlueToStopAt) {
-				rightGluePos = i;
+			var cmd = (obj instanceof ControlCommand) ? obj : null;
```
ephread: Instead of specifying the type in a comment above and performing the test in the condition, I perform the test inline and assign accordingly, which mimics what the C# code does with `as`.

Let me know if should revert the change!

engine/StoryState.js
```diff
@@ -597,49 +589,71 @@ export class StoryState{
...
+		// Debug.Assert (callStack.currentElement.type == PushPopType.Function);
+		var functionStartPoint = this.callStack.currentElement.functionStartInOutputStream;
+
+		if (functionStartPoint == -1) {
+			functionStartPoint = 0;
+		}
+
+		for (var i = this._outputStream.length - 1; i >= functionStartPoint; i--) {
+			var obj = this._outputStream[i];
+			var txt = (obj instanceof StringValue) ? obj : null;
```
ephread: Same as above (Line 556).

ephread:
Just a heads up about the PR, as I've completed the port of `Story.cs`'s changes, I'm currently re-running the tests and fixing stuff as they break. I've already fixed a number of typos in `StoryState.js` and I'll update the PR when I'll be finished. \o/

engine/StoryState.js
```diff
@@ -729,30 +742,31 @@ export class StoryState{
...
 		copy._currentChoices.push.apply(copy._currentChoices, this._currentChoices);
 
 		if (this.hasError) {
 			copy.currentErrors = [];
 			copy.currentErrors.push.apply(copy.currentErrors, this.currentErrors);
 		}
 
+		if (this.hasWarning) {
+			copy.currentWarnings = [];
+			copy.currentWarnings.push.apply(copy.currentWarnings, this.currentWarnings);
```
y-lohse: Thanks for sticking to the existing code style, but boy does that make me doubt my own sanity. A simple `copy. currentWarnings = this.currentWarnings.slice()` would have the same effect as far as I can tell?

engine/StoryState.js
y-lohse: No that's actually way better, good idea 👍

engine/StoryState.js
```diff
@@ -312,31 +299,45 @@ export class StoryState{
...
-				return g;
-			} else if (c instanceof ControlCommand) // e.g. BeginString
-				break;
+
+	CleanOutputWhitespace (str){
+		var sb = new StringBuilder(str.Length);
+
+		var currentWhitespaceStart = -1;
+
+		for (var i = 0; i < str.Length; i++) {
```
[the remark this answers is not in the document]
y-lohse: Argh, one of my comments didn't make it. There's another `str.Length` a few lines above.

engine/StoryState.js
y-lohse: Nope that's actually way better, good idea 👍

y-lohse:
By the way, feel free to start the PR for `Story.js`, or add it in this one if it's easier. I have a long train ride coming up on saturday so I'll have plenty of time to review things :)

engine/StoryState.js
```diff
@@ -187,35 +179,33 @@ export class StoryState{
...
-	}
-	set currentPath(value){
-		if (value != null)
-			this.currentContentObject = this.story.ContentAtPath(value);
 		else
-			this.currentContentObject = null;
+			return pointer.path.toString();
+	}
+	get currentPointer(){
+		return this.callStack.currentElement.currentPointer.copy();
```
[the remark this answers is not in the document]
ephread: I took an extremely cautious approach, adding `copy()` everywhere I could think of and, even then, I still managed to shoot myself in the foot. I spend the last few hours debugging an infinite loop, only to discover that `Pointer.Null` wasn't returning a brand new object every time. Stupid, stupid oversight 😂.

As soon as all the test are passing again, we'll probably want to make a set of rules on how to handle value semantics in projects (i. e. which function has the responsibility of making the copy.)

ephread:
Thanks for taking a look!

Most of the typos have now been fixed (thanks to the tests), but I'll updatethe PR later.

Story is massive, you probably want a separate PR. :) I'll also make an another PR which will fix a number of issues found in the other files I've updated. I'll try to keep everything clean and avoid a somewhat messed up history.

ephread:
> I have a long train ride coming up on saturday so I'll have plenty of time to review things :)

Some monstrous diffs are waiting for your train ride! (#171, #172)

PR #171, opened 2018-07-07
feat: update Story with all the massive change brought by v8

ephread:
Massive diff incoming!


ephread:
> My brain is fried.

Haha, mine is too, after trying to get my head around why some tests refuse to pass. I've read the entire diff again to make sure that I hadn't missed anything. Can't find the culprit!

PR #172, opened 2018-07-07
fix: correct typos and indentation on multiple files

ephread:
This PR fixes a number of small typos that I've encountered while running the tests.

Merging it after #171 might introduce some mess in the history, so I'll rework the PR if that's the case.

Side note: I've added a `copy` method to `SearchResult` in order to highlight the fact that it's a value-type. It's not used anywhere.

### DIFF
--- a/engine/Object.js
+++ b/engine/Object.js
@@ -49,21 +49,21 @@ export class Object{
 		}
 		return ancestor;
 	}
-	
+
 	ResolvePath(path){
 		if (path.isRelative) {
 			var nearestContainer = this;
 
 			if (nearestContainer instanceof Container === false) {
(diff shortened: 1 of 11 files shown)


engine/Object.js
```diff
@@ -106,22 +106,22 @@ export class Object{
...
-		} 
+		}
 		else {
 			var relativePath = this.ConvertPathToRelative(otherPath);
 			relativePathStr = relativePath.componentsString;
 			globalPathStr = otherPath.componentsString;
 		}
 
-		if (relativePathStr.Length < globalPathStr.Length) 
+		if (relativePathStr.length < globalPathStr.length)
```
y-lohse: In hindsight, `path.length` should probably be `path.Length`. Let's keep it like that and fix it later.

PR #183, opened 2018-07-19
Switching to typescript

y-lohse:
Continuing #175!

- [x] CallStack.js
- [x] Choice.js
- [x] ChoicePoint.js
- [x] Container.js
- [x] ControlCommand.js
- [x] Divert.js
- [x] Glue.js
- [x] InkList.ts
- [x] JsonSerialisation.js
- [x] ListDefinition.ts
- [x] ListDefinitionsOrigin.ts
- [x] NativeFunctionCall.js
- [x] Object.js
- [x] PRNG.ts
- [x] Path.ts
- [x] Pointer.js
- [x] PushPop.ts
- [x] SearchResult.js
- [x] StopWatch.ts
- [x] Story.js
- [x] StoryException.ts
- [x] StoryState.js
- [x] StringBuilder.ts
- [x] Tag.ts
- [x] Value.js
- [x] VariableAssignment.ts
- [x] VariableReference.js
- [x] VariablesState.js
- [x] Void.js

Stuff to review (from https://github.com/y-lohse/inkjs/pull/175#issuecomment-405284159 & https://github.com/y-lohse/inkjs/pull/175#issuecomment-405838958)

#### Checks
- [x] Check that all `Object` have been converted to `Map` where appropriate (and that it works properly, especially in `InkList`)
- [x] Check Null Exception handling
- [x] Check that `const` are only used for constants.
- [ ] Check all type assignments in conditionals.
- [ ] Check that all `any` make sense (and are not placeholders for other types).
- [ ] Check access modifiers.
- [x] Check handling of ES6 proxies in `VariablesState`.
- [x] Check all TODOs ([#194 (comment)](https://github.com/y-lohse/inkjs/pull/194#issue-205120277)).

#### Cleanup & style
- [ ] Turn C# _params_ into TypeScript _rest parameters_.
- [x] Remove all original C# comments, keeping only those pertaining to the JavaScript code.
- [ ] Reorder members according to the original order, in all files.
- [ ] Convert C#'s `as` into `asOrNull` and `(type)` into `hasOrThrow` where applicable.
- [x] Convert `forEach` loops into `for...of` loops.

#### Decisions
- [x] Raising exception vs. having undefined behavior (related to `JsonSerialisation` and `Value`, e. g. `(int) variable` vs `parseInt(variable)`). -> #227
- [x] Explicitly annotating the return type vs. letting TypeScript infer it. **-> inferring**
- [x] Use of _TryGet_ functions. -> #222 
- [x] Use of _throwNullException_ function.
- [x] Porting all `Debug.Assert` from C#. -> #221 
- [x] Define the rules about where initialize properties. **-> following the C# order as much as possible**
- [x] Decide what we do about the [`unified-signature` rule](https://github.com/y-lohse/inkjs/pull/189#issuecomment-407998700)

### Changes from master since the port has started

https://github.com/y-lohse/inkjs/compare/297f0b2e6a91d728c48a93034ebfddfba6fc4983...master

- [x] https://github.com/y-lohse/inkjs/commit/fb20bcf173e2dbd624a3740e78e81fcc5cda2cf1
- [x] https://github.com/y-lohse/inkjs/commit/b13f723237c923edacb4535d8674c8bf597bb526
- [x] https://github.com/y-lohse/inkjs/commit/50811a85522f7fef1f4d9fb1d9e069fc59d7803e
- [x] https://github.com/y-lohse/inkjs/commit/333854e920965a8ceecd7b0b3816bb6f10666e28
- [x] https://github.com/y-lohse/inkjs/commit/b1114040700589af2a66e27aa5cb5c23d32c8108
- [x] https://github.com/y-lohse/inkjs/commit/1b79e486d03ee4225322b2c646bb8fa0e78a9c4a
- [x] https://github.com/y-lohse/inkjs/commit/00d9123dc1b2755eded27b1dc9d4488024e425cd
- [x] https://github.com/y-lohse/inkjs/commit/91b038ac77970392f186fa7e80cfdbe6eb5a75ec


ephread:
@y-lohse `Void`, `ControlCommand`, `Glue`, `VariableReference` have been done in 
0abef65 by the way!

ephread:
So, `Value` is the next bottleneck as it needs to be converted before tackling `NativeFunctionCall` and `VariableStates`.

@y-lohse would you be able to add something like _Member order and property init style_ to **Stuff to review** (as per #161, and [ephread#1](https://github.com/ephread/inkjs/pull/1))?

ephread:
No worries, wasn't trying to pressure you, though. It was just in case someone would jump in to help!

ephread:
Now that `Value` is completed, who wants to do what? Let's synchronize @NQNStudios, @y-lohse!

NQNStudios:
VariablesState and StoryState look the most interesting to me, in terms of better understanding the flow of Ink. StoryState depends on CallStack, though, so maybe it makes sense for me to start with VariablesState while someone else does CallStack.

y-lohse:
Sounds good to me! @ephread what do you feel like tackling next? I might take NativeFunctionCall, but I don't really mind either way.

ephread:
@y-lohse I don't mind either way, I guess I'm going to do `Callstack` then!

y-lohse:
Since we're almost done with the actual porting, I'll kick off discussions about the decisions. Let's try it all in this PR, we'll see if we need to split up discussions.

### 1. Raising exception vs. having undefined behavior (related to JsonSerialisation and Value, e. g. (int) variable vs parseInt(variable)).

I'm not sure what this one is about, can we have a bit more details?

### 2. Explicitly annotating the return type vs. letting TypeScript infer it.

Is there any upside to define the return types explicitly when TS can infer it? It looks like unnecessary extra work to me.

### 3. Use of TryGet functions.

I have to look into that one before I have an opinion myself, but if you have one, shoot!

### 4. Use of throwNullException function.

Nothing to complain about here — it highlights places were variables could be null, and makes the errors more explicit should they happen. The impact on readability is very manageable.

### 5. Porting all Debug.Assert from C#.

Again, i have to look into that one. But it looks like something that can be done afterwards? It also ties in with the use of the `console` in general.

### 6. Define the rules about where to initialize properties.

As in, whether we define them inside vs outside the constructor?

If so, I'm very much in favor of declaring them outside, specifically having the declaration order match the C# order.

### 7. Decide what we do about the unified-signature rule

[So this is the base rule](https://palantir.github.io/tslint/rules/unified-signatures/): 

> Warns for any two overloads that could be unified into one by using a union or an optional/rest parameter.

I suspect that forcing the unification will actually make declarations less readable. Am I the only one? Whereas I really don't mind the multiple declaration.

ephread:
> 1. Raising exception vs. having undefined behavior (related to JsonSerialisation and Value, e. g. (int) variable vs parseInt(variable)).
> 
> I'm not sure what this one is about, can we have a bit more details?

For instance, `(int) value` will raise an exception while `parseInt()` will always return a value, albeit an incorrect one. (see [here (JS)](https://github.com/y-lohse/inkjs/pull/205/files#diff-e75545003680523ecc43e42f4e59f215R211) and [here (C#)](https://github.com/inkle/ink/blob/0.8.1/ink-engine-runtime/JsonSerialisation.cs#L244))

> 2. Explicitly annotating the return type vs. letting TypeScript infer it.
> 
> Is there any upside to define the return types explicitly when TS can infer it? It looks like unnecessary extra work to me.

I concur.

> 3. Use of TryGet functions.
>
> I have to look into that one before I have an opinion myself, but if you have one, shoot!

Opening another issue on this topic may be a good idea!

> 4. Use of throwNullException function.
> 
> Nothing to complain about here — it highlights places were variables could be null, and makes the errors more explicit should they happen. The impact on readability is very manageable.

I'd like to make one change: removing the string parameter from `throwNullException`. Ideally, we'll update the test and get proper source maps, which will make tracking issue easier.

> 5. Porting all Debug.Assert from C#.
> 
> Again, i have to look into that one. But it looks like something that can be done afterwards? It also ties in with the use of the console in general.

Definitely, we could open another issue to deal with this. I kind of like the idea of being explicit, but maybe we should disable any console use in production (using a flag?).

> 6. Define the rules about where to initialize properties.
>
> As in, whether we define them inside vs outside the constructor?
>
> If so, I'm very much in favor of declaring them outside, specifically having the declaration order match the C# order.

That's correct. I think we should follow strictly what the C# does, and in the uninitialized properties, initialize them outside of the constructor.

y-lohse:
Created #221 and #222 to track decisions decisions that can be postponed.

y-lohse:
@NQNStudios I think you ere the one commenting on proxies [here](https://github.com/y-lohse/inkjs/blob/typescript/src/VariablesState.ts#L94)? If so, do you remember what you had in mind?

src/Container.ts
```diff
@@ -235,10 +235,10 @@ export class Container extends InkObject implements INamedContent{
 
 		sb = arguments[0] as StringBuilder;
 		let indentation = arguments[1] as number;
-		const pointedObj = arguments[2] as InkObject | null;
+		let pointedObj = arguments[2] as InkObject | null;
 
 		function appendIndentation(){
-			const spacesPerIndent = 4; // Truly const in the original code
+			let spacesPerIndent = 4; // Truly let in the original code
```
ephread: "Truly let" 😂, you dumb computer.

y-lohse: Woops, good catch^^

y-lohse:
We're getting there! Most of the remaining checklist are non-urgent things.

Just a few more questions for you @ephread :

1. There are some `TODO` markers in [CallStack](https://github.com/y-lohse/inkjs/blob/typescript/src/CallStack.ts#L71) that look like dev hints. Is there actually something to do on them?
2. You mentioned wanting to remove the parameter from `throwNullException`. Do you want to do this as part of this PR or at some later point?
3. [What is this one about?](https://github.com/y-lohse/inkjs/blob/typescript/src/JsonSerialisation.ts#L279)

ephread:
> We're getting there! Most of the remaining checklist are non-urgent things.

Awesome!

> There are some TODO markers in CallStack that look like dev hints. Is there actually something to do on them?

It's a reminder to decide whether we should throw if the value is not of the correct type (to match the C# code). What do you think?

> You mentioned wanting to remove the parameter from throwNullException. Do you want to do this as part of this PR or at some later point?

It can wait, as we need to refactor the tests first (to get the proper line report when the test fail). At the moment, the test suit run against the compiled & concatenated JavaScript file and doesn't pick up the source maps, resulting in funky line numbers.

> What is this one about?

I believe it's a reminder to make sure we that check against the appropriate value. `undefined` might make more sense here, since the _absence of value_ will be `undefined`, rather than C#'s `null`.

y-lohse:
> It's a reminder to decide whether we should throw if the value is not of the correct type (to match the C# code). What do you think?

Ah, ok! [I opened an issue for this](https://github.com/y-lohse/inkjs/issues/227), but I think we should throw. Since there's an issue for that I'll consider it non blocking for the merge.

> It can wait, as we need to refactor the tests first (to get the proper line report when the test fail). At the moment, the test suit run against the compiled & concatenated JavaScript file and doesn't pick up the source maps, resulting in funky line numbers.

Alright!

> I believe it's a reminder to make sure we that check against the appropriate value. undefined might make more sense here, since the absence of value will be undefined, rather than C#'s null.

Ah yes, good catch!

y-lohse:
Alright, it's time to merge this PR!

First of all I want to thank @ephread for leading this effort and doing the bulk of the work. Big thanks as well to @NQNStudios for participating in this effort.

The remaining items on the checklist are not blocking and can be sorted over time. Some already have dedicated issues, other will get them as we go. Keeping this PR open until everything is checked would drag things out forever.

Going forward, I'll focus on making sure the changes from ink v0.8.2 get integrated before releasing a new stable version, and then avoiding having to do huge changes in one go as much as possible.

PR #184, opened 2018-07-19
Typescripting Pointer

y-lohse:


src/Pointer.ts
```diff
@@ -0,0 +1,56 @@
...
+		return this.container == null;
+	}
+
+	get path(): Path | null{
+		if (this.isNull) return null;
+
+		if (this.index >= 0)
+			return this.container.path.PathByAppendingComponent(new Path.Component(this.index));
+		else
+			return this.container.path;
```
y-lohse: @ephread I'm afraid i need a tp on this again!

The compiler complains that `this.container` can be `null`. it looks like it doesn't consider `if (this.isNull) return null;` as a valid exit path, presumably because the actual variable comparing is not inline.

Is there a way to tell the compiler that it's fine, we're guarded against being `null`already? I was looking at typeguard annotations, but they only seem to allow informations such as `container is Container`, not `container is not null`.

I guess if it comes to the worth I can always inline the check, but is there a better way to handle this?

ephread: > Is there a way to tell the compiler that it's fine, we're guarded against being nullalready?

Yes! Use [`!`](type-guards-and-type-assertions). If you're 100% certain that the execution path won't be reached if a particular element is `null` or `undefined`, you can force the compiler to accept the statement. That would look like:

```typescript
this.container!.path
```

---

> I was looking at typeguard annotations, but they only seem to allow informations such as container is Container, not container is not null.

Actually, (take that with a pinch of salt, I haven't double checked) since we're working with `strictNullChecks` enabled, `container is Container` also means that `container is not null`. `null` and `undefined` are treated as their own types rather than the absence of value (that's why we have unions like `Path | null`).

In our case, however, since we need to checks the inners of a class, I don't think that type guards could really be helpful. We wouldn't be able to return a type predicate pertaining to the properties.

y-lohse: Thank you! I have yet to read the typescript docs cover to cover, and I find it a bit hard to find answers to precise questions there :-)

PR #188, opened 2018-07-22
Fixing the type check for numbers from JSON

### LINKED ISSUE #179: Better number detection from JSON
Would it be possible to replace [this line](https://github.com/ephread/inkjs/blob/master/engine/JsonSerialisation.js#L78-L79) with a simple `typeof token === "number"`? Probably yes.

NQNStudios:
Fixes #179


y-lohse:
Awesome, thank you!

PR #189, opened 2018-07-23
Porting Value to typescript

y-lohse:
Yeah ok I'm not finishing this tonight, but here's a work in progress.

I used a lot of `any` so far that I can hopefully convert later. And lots of other tricky things. Diving in at the deep end, wooo!


NQNStudios:
Would it make sense for me to try and finish this off?

NQNStudios:
Maybe we could open a Gitter channel or something to talk about porting details in real-time. I'm slowly figuring things out but it would be nice to have someone walk me through a few things about the process.

joethephish:
We we have a Discord server with an ink channel, I’d also be happy to add an inkDev channel that could include discussions about the inkjs port? https://discord.gg/MUXj7Md

ephread:
@joethephish I fully support that idea! I've got a couple of questions to ask about the development of ink by the way, but I don't think these questions would warrant opening an issue.

We also definitely need to write a CONTRIBUTING.md, to detail all the best practices and rules about the transpiling process.

joethephish:
Okay, #ink-dev is open! https://discord.gg/QYs6muQ

ephread:
Great to see #190 merged! @y-lohse do you want me to review the result?

y-lohse:
Yep, I just did and made a few changes that seemed sensible :  17cf7f5, 63ec475 & 610fd1d . 
Would you and @NQNStudios mind also taking a look at it? Once everything is ok I'll clean up the commit history and merge it.

I think we should disable the `tslint:disable:unified-signatures` since it's now a comment in at least 2 places?
Also the linting currently fails because of the no-assignments in conditions rule. @ephread have you seen that case before? The C# version uses them a lot so I'd be inclined to disable that rule as well, even though I'm not super comfortable with assigning in conditions myself.

ephread:
> I just did my own review and made a few changes that seemed sensible : 17cf7f5, 63ec475 & 610fd1d .

Great, thanks!

> I think we should disable the tslint:disable:unified-signatures since it's now a comment in at least 2 places?

I'd be enclined to keep the rule enabled for now, since it makes sense in the context of method overloading. In the case of overloaded constructors, however, I originally disabled the rule so that the overloads and their C# counterparts would share a fairly similar signature. Let's review these rules at the end of the conversion.

> Also the linting currently fails because of the no-assignments in conditions rule. @ephread have you seen that case before? The C# version uses them a lot so I'd be inclined to disable that rule as well, even though I'm not super comfortable with assigning in conditions myself.

I don't like assigning within conditions either. I'm not sure those assignments happen that much in the C# codebase though. In 0.8.1, I couldn't really find occurrences of them within `if` constructs. But I know that they have occurred in JavaScript and I believe they stem from the `TryXXX` constructs, which were transpiled in different ways, in different locations, depending on the types involved.

In an effort to normalize the construct, I've created the `TryGetResult` interface, based on previous efforts. We should probably add two functions `TryParseInt` and `TryParseFloat` which would return a `TryGetResult` (renaming it to `TryResult` is probaby a good idea) and [test the exists property in the `if`](https://github.com/y-lohse/inkjs/blob/typescript/src/InkList.ts#L160).

Could look like this:

```typescript
export function tryParseInt(value: any, /* out */ defaultValue: number): TryGetResult<number> {
	let val = parseInt(value);

	if (isNaN(val)) {
		return { result: val, exists: true };
	} else {
		return { result: defaultValue, exists: false };
	}
}
```

I'll take a look at the diff soon!

src/Value.ts
```diff
@@ -0,0 +1,327 @@
...
+	public toString(): string{
+		return 'DivertTargetValue(' + this.targetPath + ')';
+	}
+}
+
+export class VariablePointerValue extends Value<string>{
+	public _contextIndex: number;
+
+	constructor(variableName: string, contextIndex: number = -1){
+		super(variableName);
```
ephread: Here, the original C# code has an empty constructor which initialize `value` to null (through super). Don't / Do we want this? (See the top comment as well.)

y-lohse:
Thanks for the review!

### `unified-signature`

Works for me! Added to the end-of-conversion checklist.

### Assignments in conditions and `TryGetValue`

Ah yes, all the assignments in conditions are probably `outs` or casts, which we have ways to deal with now. So the rule should stay, and the code should change.
I'm not too sure about returning an object with an `exists` flag all the time. But I have to look into it properly, re-read the rational behind it etc. So another thing for the end-of-conversion, and in the meantime let's do as you suggest.

### Contructors

Yep, I think that makes sense. So in the case of `Value`, it only needs to match the definition order, right? Or should the signature of the catch-all function change?

### Return types

I thought specifying them was mandatory, but no. So yeah, let's remove them whenever they aren't needed.

src/Value.ts
y-lohse: Yeah I didn't change it because lazy (if `Value.value` can be `null` we need to add a bunch of safeguards), but we should stick to the C# behavior here as well.

ephread:
LGTM! We should also probably add _turning all `.forEach` into `for...of`_ to the checklist.

> So in the case of Value, it only needs to match the definition order, right? Or should the signature of the catch-all function change?

It depends on your taste, honestly.

When there are a lot of different constructors, I like to have a catch-all with no parameters and use the `arguments` array instead. That way, the signatures are almost the same and you don't get parameters with union types. While it makes the catch-all constructor _ a bit less_ readable†, it make the signatures more C#-like.

But, ultimately, it's a matter of taste. That's something we should normalize later on.

†(Since you will manually coerce `argument[x]` to the appropriate type and assign the resulting value to a variable with the same name as the parameter.)

y-lohse:
Ok cool, merging this then! Nice piece of team work :)

PR #190, opened 2018-07-24
Finished porting Value to Typescript

NQNStudios:
Continuing @y-lohse's work on #189, I believe I've successfully finished porting Value.js to Value.ts.


y-lohse:
Thank you! I'll take a look at this in about 10 hours.

We'll have to rebase this because I think we comited more or less at the same time, but your changes are more advanced. I can take care of it tonight if no one has done it until then.

I'll have a bunch of TS-related questions and review comments, but I think it's easier if we do this on the other PR to avoid scattering the conversation :)

PR #193, opened 2018-07-31
WIP porting VariablesState

NQNStudios:
Here's the result of today's work by @elabraha and myself on VariablesState. As it is, almost every test fails. We are not 100% sure why. Will work on it more tomorrow but if anyone wants to take a look, here you go :P


y-lohse:
Closing this one since it's taken over by #210 !

PR #194, opened 2018-07-31
feat: convert CallStack

ephread:
I've reordered the members to match the order found in the original C# file. I've also added a number of `TODO`s, we'll need to review them at the end!


y-lohse:
Thanks! The conversion looks good to me, but we need to [export the `Thread` type again](https://travis-ci.org/y-lohse/inkjs/builds/410335800#L874)

ephread:
🤦 I forgot to rebuild before running the tests locally, hence they ran against the last version. I'll fix all the issue!

PR #195, opened 2018-07-31
Convert NativeFunctionCall

y-lohse:
Still lots to do.


y-lohse:
Ok sorry for the delay, but I struggle to find time for this, and the remaining problems I'm facing need some research. Any insight or pointers in the right direction would be appreciated!

1. I have a problem with [this](https://github.com/inkle/ink/blob/master/ink-engine-runtime/NativeFunctionCall.cs#L126) signature. It doesn't look like TS lets me have a `Call` and `Call<T>` definition? It also really wants `parametersOfSingleType` to be a specific `Value` type, but it should be the abstract Value.
2. [I want to cast stuff as `Value` of a type `T`](https://github.com/inkle/ink/blob/master/ink-engine-runtime/NativeFunctionCall.cs#L131). Can I do that with `asOrThrows`?
3. I also want to cast variables as `Value`, but again TS won't let that fly because `Value` is abstract. What's the recommended way around that?

Let's start with these questions, hopefully I'll be able to figure out the rest by myself.

ephread:
> Ok sorry for the delay

No worries!

You probably picked one of the most tricky file, by the way. Not necessarily the most tedious (400 lines are nothing compared to 1800 lines – looking at you `Story` ☠️), but from an intellectual standpoint, it's definitely more challenging than `Story`.

> I have a problem with this signature. It doesn't look like TS lets me have a Call and Call<T> definition?

That's because it works like the constructors. A couple of signatures, followed by a non-overloaded, catch-all function. So you're going to have to write something like:

```typescript
public Call(parameters: InkObject[]): InkObject | null;
public Call<T>(parametersOfSingleType: Array<Value<T>>): Value<T>
public Call<T>(){
    if (/* InkObject[] */) {

    } else if (/* Array<Value<any>> */) {

    }
}
```

We'll need a way to discriminate, at runtime, between the two possible arguments (which are both arrays), checking the type of the first element of the array might be a way.

> It also really wants parametersOfSingleType to be a specific Value type, but it should be the abstract Value.

I think you can use `Value<any>` to get the same behavior as `Value`, since it'll match _any kind_ of `Value`. Yes, it seems that TypeScript won't let you specify an non-parametrized generic type in the signature, but I don't know the reasons.

That said, If I understand the code correctly, it seems that what we really want is actually `Value<T>`, which would make `(Value<T>)param1` unneeded.

> I want to cast stuff as Value of a type T. Can I do that with asOrThrows?

Sadly, no, because generics only exists in TypeScript and are not translated into Javascript. It means that `Value<T>` is purely virtual and is not a valid type that you can pass to a function. This is probably one of those edge cases where we'll have to use an _unsafe_ type assertion like `var val1 = param1 as Value<T>`.

It's the same for `BinaryOp`:

```typescript
let intOp = this._operationFuncs.get(ValueType.Int) as BinaryOp<number>;
```

I'm not sure we can come up with a satisfying solution to check these types at runtime. I believe it's alright, though, since we'll be calling JavaScript functions on the other end anyway. It'll work its dark magic and types will most likely be checked by hand, in true JavaScript fashion.

> I also want to cast variables as Value, but again TS won't let that fly because Value is abstract. What's the recommended way around that?

Actually, a while ago, I've updated the method in one of the opened PR. So you can now pass abstract classes (which are of type: `Function & { prototype: T }`). See [here](https://github.com/y-lohse/inkjs/pull/206/files#diff-af457ce62ed297249d7f001fb51d69e2R13).

---

By the way, we'll need to pick a syntax for delegates, basically:

```typescript
export type BinaryOp<T> = (left: T, right: T) => any;
export type UnaryOp<T> = (val: T) => any;
```

vs.

```typescript
interface BinaryOp<T> {
	(left: T, right: T): any;
}
interface UnaryOp<T> {
	(val: T): any;
}
```

I feel that the first option (also preferred by the linter) is a bit more easy to understand, unless we're JavaScript specialists. That's another thing to add the list of checks!

y-lohse:
Yay it's green! Thank fou for the tips, they really helped.

I should probably re-read the whole jazz again because some things might be unfinished, and a fresh set of eyes on it would probably do good.

And then i think we can start the Big Merge® \o/

ephread:
@y-lohse You're welcome!

> And then i think we can start the Big Merge® \o/

That's awesome! I can review this PR whenever you need me to.

y-lohse:
Ok it's ready for review now :)

I left `Call` and `Call<T>` as separate functions, because we can't infer which one to call based on the type of parameters — they both get possibly called with arrays of `Values`, which are `InkObjects` too. I don't think it's a problem though because one of them is actually private.

Other than that nothing special to note I think.

src/NativeFunctionCall.ts
```diff
@@ -0,0 +1,436 @@
...
+			return this._prototype.numberOfParameters;
+		} else {
+			if (this._numberOfParameters === null) return throwNullException('NativeFunctionCall._numberOfParameters');
+			return this._numberOfParameters;
+		}
+	}
+	set numberOfParameters(value: number){
+		this._numberOfParameters = value;
+	}
+	public _numberOfParameters: number | null = null;
```
ephread: I'm not sure this should be allowed to be `null`, since in the original code the default value will be 0. Did you run into issues with a non nullable `_numberOfParameters`?

src/NativeFunctionCall.ts
```diff
@@ -0,0 +1,436 @@
...
+
+	public CallBinaryListOperation(parameters: InkObject[]){
+		if ((this.name == '+' || this.name == '-') && parameters[0] instanceof ListValue && parameters[1] instanceof IntValue)
+			return this.CallListIncrementOperation(parameters);
+
+		let v1 = asOrNull(parameters[0], Value);
+		let v2 = asOrNull(parameters[1], Value);
+
+		if (v1 === null) return throwNullException('NativeFunctionCall.CallBinaryListOperation v1');
+		if (v2 === null) return throwNullException('NativeFunctionCall.CallBinaryListOperation v2');
```
ephread: We should probably use `asOrThrows` directly here, instead of `asOrNull` followed by a manual null check. It does make the code slightly different, but not by much. Nat did something similar in VariablesStates and I found the change to be more elegant in the end.

As long as we explain the rationals in a CONTRIBUTING.md, we'll be fine, what do you think?

y-lohse: Sounds good!

src/NativeFunctionCall.ts
y-lohse: Nah I thought the default value was `null`, not `0`. i'll change that!

PR #201, opened 2018-08-15
Typo fixes in callstack.js (fixes #200)

joethephish:

PR #204, opened 2018-08-20
Convert StoryState

ephread:
As usual, I took the chance to clean some things up, so I replaced certain constructs with others and removed two methods which were no longer used. The fixes from #201 have been included as well.

#### Changes in helpers

1. The `JObject` type, which is basically a regular Javascript Object, but with a type signature, so that TypeScript is happy.
2. `nullIfUndefined` which basically returns `null` if an object is undefined (useful since in Javascript, `Map.get` returns `undefined` if a key does not exists, while the ported code expect `null` values.)
3. `asOr` functions have been updated to support abstract classes; the type union is probably unnecessary but I'll check that later.

Maybe it would be a good idea to gather all Javascript-only helpers in their own subdirectory.


ephread:
PR updated and ready, thanks for reviewing!

PR #205, opened 2018-08-20
[Do Not Merge] [WIP] Json Serialisation

ephread:
**[Do Not Merge] This PR contains commits from #195, [ephread/VariableState](https://github.com/ephread/inkjs/commits/VariablesState) and #204. Once all the previous changes are merged, I'll rebase this PR to have a clean merge**.

b1114040 was also ported.

The port is more or less completed, except for a few choices that we need to discuss.

1. I've disabled the `no-conditional-assignment` linting rule in the file. As per [#189-407986191](https://github.com/y-lohse/inkjs/pull/189#issuecomment-407986191), conditional assignments are not desirable, but how should we replace them? When accessing plain old JavaScript object it might be a bit overkill to involve a new `TryGet` method, but, on the other hand, dynamically accessing a property more than once is definitely typo prone. Extending the prototype is probably not a good idea either.

2. I've filtered the `for...in` constructs using `hasOwnProperty`. It's a safe addition but it may also be a useless one, given the type of objects we're working with.


ephread:
Ready for review!

PR #206, opened 2018-08-23
Story

ephread:
This is a big file. Comments should help clear out any confusion during the review, but feel free to ask if some constructs are unclear.

### DIFF
--- a/src/JsonSerialisation.ts
+++ b/src/JsonSerialisation.ts
@@ -383,7 +383,7 @@ export class JsonSerialisation{
 		// var controlCmd = obj as ControlCommand;
 		let controlCmd = asOrNull(obj, ControlCommand);
 		if (controlCmd) {
-			return JsonSerialisation._controlCommandNames[Math.floor(controlCmd.commandType)];
+			return JsonSerialisation._controlCommandNames[controlCmd.commandType];
 		}
 
 		// var nativeFunc = obj as Runtime.NativeFunctionCall;


ephread:
Thanks for reviewing!

src/JsonSerialisation.ts
```diff
@@ -383,7 +383,7 @@ export class JsonSerialisation{
 		// var controlCmd = obj as ControlCommand;
 		let controlCmd = asOrNull(obj, ControlCommand);
 		if (controlCmd) {
-			return JsonSerialisation._controlCommandNames[controlCmd.commandType];
+			return JsonSerialisation._controlCommandNames[Math.floor(controlCmd.commandType)];
```
y-lohse: Is the `Math.floor` call there to cast the value as an int?

ephread: I… think so. Honestly, upon further inspection it's absolutely useless, since `ControlCommand` is an `enum`. That's a good catch, let me remove this.

ephread:
> Wow, +1,984 −1,757… thank you so much for the effort, that must have been hardcore.

Haha, you're welcome. That wasn't too bad, it was far less nerve-racking than porting the 0.7.5...0.8.1 diff.

> Is there anything that would be worth reviewing in particular? Otherwise I'll go ahead and merge this PR too.

Three particular bits come to mind, since they are quite error-prone.

1. [Story's constructor](https://github.com/y-lohse/inkjs/pull/206/files#diff-f23e4d82357ae59f32fe9152b3a89efaR107);
2. [ValidateExternalBindings](https://github.com/y-lohse/inkjs/pull/206/files#diff-f23e4d82357ae59f32fe9152b3a89efaR1405);
3. Anything related to `_variableObservers`.

src/Story.ts
```diff
@@ -0,0 +1,1888 @@
...
+			} else {
+				json = arguments[0] as JObject;
+			}
+		}
+
+		// ------ Story (Container contentContainer, List<Runtime.ListDefinition> lists = null)
+		if (lists != null)
+			this._listDefinitions = new ListDefinitionsOrigin(lists);
+
+		this._externals = new Map();
```
y-lohse: I don't think it matters too much because we already did it in the js version, but it looks like `_externals` should [only be initialized](https://github.com/inkle/ink/blob/master/ink-engine-runtime/Story.cs#L155) in the `Container` version of the constructor. Do you think we should change that?

ephread: True, but it turns out that [`Story(string)` calls `Story(Container, List<RUntime.ListDefinition>?)`](https://github.com/inkle/ink/blob/master/ink-engine-runtime/Story.cs#L161), so _externals will always be initialized (or I am reading this wrong?).

The constructor is a bit confusing as it is. Maybe we could do a better job, but I'm not sure. Still, I'm going to move L123 out of the argument validation block, so that it'll look more akin to the original code.

PR #210, opened 2018-08-30
Variables State

ephread:
This PR completes #193.

Thanks for reviewing!

### DIFF
--- a/src/VariablesState.ts
+++ b/src/VariablesState.ts
@@ -1,12 +1,10 @@
-//still needs:
-// - varchanged events
-// - see if the internal getenumarators are needed
-import {Value, VariablePointerValue, ListValue} from './Value';
+import {AbstractValue, Value, VariablePointerValue, ListValue} from './Value';
 import {VariableAssignment} from './VariableAssignment';
 import {InkObject} from './Object';
 import {ListDefinitionsOrigin} from './ListDefinitionsOrigin';
 import {StoryException} from './StoryException';
 import {JsonSerialisation} from './JsonSerialisation';
+import {asOrThrows, asOrNull} from './TypeAssertion';
 
 export class VariablesState{
     private _globalVariables: Map<string, InkObject>;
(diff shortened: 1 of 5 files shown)


src/VariablesState.ts
```diff
@@ -0,0 +1,311 @@
...
+		this.variableChangedEventCallbacks.forEach((callback) => {
+			callback(variableName, newValue);
+		});
+	}
+
+	get batchObservingVariableChanges(){
+		return this._batchObservingVariableChanges;
+	}
+	set batchObservingVariableChanges(value: boolean){
+		value = !!value; // TODO: Might have become useless with TypeScript.
```
y-lohse: Yup, useless now.

src/VariablesState.ts
```diff
@@ -0,0 +1,311 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	get callStack(){
+		return this._callStack;
+	}
+	set callStack(callStack: any /* CallStack */){
```
y-lohse: Why can't we use the `CallStack` type here ?

src/VariablesState.ts
```diff
@@ -0,0 +1,311 @@
...
+			this.SetGlobal(variableName, val);
+		}
+	}
+
+	constructor(callStack: any /* CallStack */, listDefsOrigin: ListDefinitionsOrigin){
+		this._globalVariables = new Map();
+		this._callStack = callStack;
+		this._listDefsOrigin = listDefsOrigin;
+
+		// TODO should we discourage the use of proxies since they're non-portable
```
y-lohse: Can you elaborate more on this please?

src/VariablesState.ts
```diff
@@ -0,0 +1,311 @@
...
+			// It first tries to access the objects own property, and if none is
+			// found it delegates the call to the $ method, defined below
+			let p = new Proxy(this, {
+				get(target: any, name){
+					return (name in target) ? target[name] : target.$(name);
+				},
+				set(target: any, name, value){
+					if (name in target) target[name] = value;
+					else target.$(name, value);
+					return true; // returning a falsy value make the trap fail
```
y-lohse: I wonder what I meant by that, but it doesn't look very sound.

src/VariablesState.ts
ephread: I don't know much about proxies, this is most likely a comment inserted by @NQNStudios, so you'd have to ask them!

src/VariablesState.ts
ephread: Definitely, I guess that Callstack wasn't merged yet when the PR was first worked on.

src/VariablesState.ts
ephread: Maybe we should rework the entire proxy thing later?

y-lohse: Yah, we'll see :)

y-lohse:
Awesome, thanks for the last changes!

PR #212, opened 2018-08-31
Prevent exponential introspection of content during ValidateExternalBindings

joethephish:
Matches C# code more closely.

### DIFF
--- a/engine/Story.js
+++ b/engine/Story.js
@@ -1355,7 +1355,9 @@ export class Story extends InkObject{
 			var c = containerOrObject;
 
 			c.content.forEach(innerContent => {
-				this.ValidateExternalBindings(innerContent, missingExternals);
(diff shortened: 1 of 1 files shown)


y-lohse:
Strange, it looks like inkjs has been missing this for a really long time… Anyway thanks for the fix!

ephread:
Looks like it's already fixed in the [TypeScript version of `Story`](https://github.com/y-lohse/inkjs/pull/206/files#diff-f23e4d82357ae59f32fe9152b3a89efaR1439) by the way. So no need to port the fix.

Thanks for tracking the issue @joethephish! I feel kinda bad that you spent time investigating, while the issue ended up being magically fixed by the TypeScript conversion.

joethephish:
No problem! Didn’t take too long to find :)

PR #218, opened 2018-09-08
[WIP] 0.8.1 -> 0.8.2

ephread:
It's still a work in progress. Two tests are failing.

The changes are not particularly nasty, so the merge can theoretically be done before cleaning up the `typescript` branch.


ephread:
So, one the test fails because of the changes made to LIST_RANGE. The test will need to be updated to match the new behavior.

The other issue seems to be a rehash of the whitespace issues we've encountered many times. I probably won't have the chance to look into it this week, so if anyone wants to take it from here, feel free to do so!

src/InkList.ts
```diff
@@ -379,6 +379,39 @@ export class InkList extends Map<SerializedInkListItem, number> {
...
+	{
+		if (this.Count == 0) return new InkList();
+
+		let ordered = this.orderedItems;
+
+		let minValue = 0;
+		let maxValue = Number.MAX_SAFE_INTEGER;
+
+		if (Number.isInteger(minBound)) {
+			minValue = minBound;
```
y-lohse: We're missing the `int` casts here and for `maxBound`.

ephread: Happy to append `as Number`, but wouldn't it be redundant? We've already checked that `minBound: any` is an integer. In C# the cast is required, to coerce `object` to `int` but in TypeScript, `any` can always be implicitly coerced.

ephread:
> Awesome, thanks! That was quite pleasant to review because of typescript and matching orders.

Happy to see that we're faster than ever!

src/InkList.ts
y-lohse: Ah yes, you're correct. All good then!

PR #223, opened 2018-09-12
Fix typo in TryGetDefaultVariableValue

joningold:
Note, I don't have a repro for ink  that fires this code-path, but this is definitely the correct fix!


y-lohse:
Had to rebase this one for some reason, did it in #226

PR #239, opened 2018-09-28
[WIP] Expose type definitions and source maps. / Improve error reporting in tests.

ephread:
The goal of the PR is to expose source maps and type definitions, which comes with a number of changes, particularly concerning the build system. Jasmine / Istanbul configurations have also been updated to take advantage of the source maps.

### Build System

The TypeScript code is compiled twice (once for distribution, once for use as a dependency), with different configurations.

#### Concatenated files

Rollup now uses four more plugins to create the distributed files.

- `uglify` is used to minify the ES5 version.
- `terser` is used to minify the ES2015 version.
- `typescript2` is used to compile the TypeScript sources.
- `sourcemaps` is used to load and merge the sourcemaps throughout the build process.

During the Rollup build:

- TypeScript modules are converted into ES6 modules;
- type definitions are not generated;
- the resulting bundle is minified;
- source maps are generated.

#### Source with type definition

During the regular build:

- TypeScript modules are converted into CommonJS modules;
- type definitions are generated;
- source maps are generated.

A new file (`main.js`) exposes `Story` and is now the entry point in `package.json`.

### Jasmine / Istanbul

1. Jasmine can now output the correct TypeScript file and the correct line number when reporting an issue. This is made possible by using `jasmine-ts-console-reporter` and making small changes in the configuration.

2. Istanbul has been updated to a newer, alpha version supporting source maps. This shouldn't be a cause for concern as it's a dev dependency.

3. The `test` task has been split into `test` and `coverage` since Jasmine has sometimes been reporting wrong line numbers when ran through Istanbul (I'll investigate).

4. The coverage is now reported on the original TypeScript files.

---

I still have to update the README to point out the differences between requiring `inkjs` or `inkjs/dist/ink`.

Thanks for taking a look!


ephread:
@y-lohse any update on this? 🤞

y-lohse:
Sorry, not yet :( I'll find some time this week!

y-lohse:
First off, apologies for the huge delay with the review :(

This all looks pretty good to me but should we really change the `main` field in package.json? People who are simply use `require('inkjs')` will now get a different result, which might lead to problems. From what I can see the `main.ts` file only exposes the `Story` class, whereas the `dist/ink.js` file also exports `InkList` for example. Am I correct with this assumption or did I miss something?

Side question — does the typescript build use babel, or is it possible to make it use babel at all?

ephread:
> First off, apologies for the huge delay with the review :(

No worries, you should see my latency on other projects, thanks for reviewing!

> This all looks pretty good to me but should we really change the main field in package.json? People who are simply use require('inkjs') will now get a different result, which might lead to problems. 

I feel strongly that we should make this breaking change. It makes more sense to expose the "vanilla" JavaScript, along with type definitions (the way it's done in most TypeScript modules available through NPM). The developers will have the most flexibility, they can import the dependency in both Javascript and TypeScript, they can transpile the code to any version of ECMAScript and they can bundle the result using any module bundler they want.

For instance, [ink-vn-engine](https://github.com/paulloz/ink-vn-engine) currently uses a [type definition written by hand](https://github.com/paulloz/ink-vn-engine/blob/master/src/%40types/inkjs/index.d.ts), this won't be necessary anymore with this PR. The project also uses Browserify to bundle the modules.

> From what I can see the main.ts file only exposes the Story class, whereas the dist/ink.js file also exports InkList for example.

Exporting only `Story` might be not be sufficient indeed. Maybe `Tag`, `Choice` and `VariableState` should be exported as well (as per the file linked above). As for `InkList` I don't know if it should be accessible from outside the module, but this warrants a more thorough search.

If the expected classes are exposed and unless people are using the entry point to import the concatenated/minified file the old fashioned way (kinda the Bower way), I'm not sure the change is _that_ breaking, though.

For those still using Bower or importing the files with NPM, the compatibility break can be solved by creating a `bower.json` in this repo or by pointing to `dist/` rather than using the main entry point. Obviously, a warning about the change in the README (and perhaps in a post install script) is in order.

> Side question — does the typescript build use babel, or is it possible to make it use babel at all?

It doesn't, TypeScript currently transpiles to ES2015, using `CommonJS` and with no polyfills involved. I don't think you can use Babel without adding another transpiling phase right after. I don't think it's necessary, though. This can be handled by the developers, depending on what they want to support.

y-lohse:
>  As for InkList I don't know if it should be accessible from outside the module, but this warrants a more thorough search.

IIRC it's actually one of the only classes that is exposed in the original ink :) But yeah let's add this one and the other you mentioned.

> This can be handled by the developers, depending on what they want to support.

Thanks for the clarification :) It was more for the internal usage — it would be cool to get rid of one of the build steps. But sounds like we'll keep rollup to build the dist versions.

_______

Regarding the main file, to clarify, I'm totally in favor of exposing the vanilla js files as well as the typescript definitions, just not necessarily by default. I didn't really say why though, so here we go:

1. People who are not familiar with build systems can't use the default version — they either need to `require` a specific file, or set-up a build system, which might be complicated for someone who's not familiar with them
2. People who already have a build system in place and require the default version will now build inkjs on top of their own files. It's not a big thing, but well.

On the other hand I think imposing the burden to `require(inkjs/engine/Story)` on devs who use typescript or explicitly want to build inkjs themselves is not a big problem?

Do you see any major pain points in having the vanilla version *not* be the default version?

y-lohse:
> Welcome to the JavaScript world! :)

Yeah how dare you.

> Does that sound good to you?

Yep, that would work for me! Thanks :)

PR #319, opened 2019-05-05
Pull to fix #318.

### LINKED ISSUE #318: Usage of instanceof in JsonSerialisation.ts fails when data is shared between windows.
Due to the way that iframes produce a new window object, and instanceof Array refers to window.Array, if an array created in an iframe is passed to the inkjs library on the top window, it will fail, because array AST objects fail the instanceof check. Switching these calls to Array.isArray should fix this issue.

I have added a pull request at #319. This fixes the incidences I have found in JsonSerialisation but it is possible there are other identical issues elsewhere. Feel free to disregard the PR if you'd rather ignore it or deal with it yourself.

furkle:
This pull should fix the issues outlined in #318.


y-lohse:
Thank you!

PR #335, opened 2019-05-31
feat: added types information

Hammster:
Hi,

Great lib you got there, sadly it had no default types on the `require('inkjs')` which is not great when working with typescript or intellisense, I've added the necessary file and change on the package.json to make this work.

It should work now with the common import systems.

CommonJS require 

Code_2019-05-31_20-48-21

ESM imports

Code_2019-05-31_20-47-52


ephread:
Hi @Hammster, thanks for you PR!

Note that full type definitions are exported from TypeScript, but they are only available if you import ink through `inkjs/engine/runtime`, e. g. : `import { Story } from "inkjs/engine/runtime"`.

See #239 for more information. The README is a bit outdated, I apologise for that.

Hammster:
Thanks for the hint 🙂 

I've read through the issue now. My point on that take is that the types are not only for TS devs it is information for any IDE and editor as well within regular JS files. Hiding the typing info within a specific import is not the natural way of the JS exosystem, to be honest I don't even know a popular framework doing so.

What I get is it would be nice to have access to type information for classes like `Tag` `Choice` `VariablesState` etc. through a proxy import like they are exposed in the `runtime.ts`. But currently this is more like having two separate libraries in one package, my proposal in that regard would be to expose these classes and their type information through the main file, it would be a single defined API instead of  two separate entry points, making this awesome lib more accessible and eco-system friendly. 🙂

ephread:
Well, per #239, I completely agree with you. I'd like to see the type definition and the non-concatenated source code exposed through `main`, instead of it pointing to `ink-es2015.js`. That way, devs can build their JavaScript bundles however they want and take advantage of the type definition.

It's a breaking change though (I think it would only break the current workflow for very few people, but I don't have any number to back that up), so we decided not to do it.

Let's way for @y-lohse to give us his thoughts on the matter.

y-lohse:
Hi, thanks for the issue and PR! I will re-read  #239 and re-evaluate if it's worth publishing a breaking change or not.

y-lohse:
Ok so: I'm happy to merge this PR as it is and do a new release, it's not a breaking change so no problem.

Regarding the other option - changing what the `main` field points too - I'd accept a PR for this too. I'm still not convinced it's for the best, but you've convinced me that people who really care about the `main` field are more advanced users that want could use the extra power. Ideally the PR for that would:

- Still produce a `ink-es2015.js` somewhere in the dist files that go to npm
- Update the "Installation" part of the doc to reflect the new ways to use inkjs
- And of course expose the various types

We can then release that under a new major version number and explain how to transition in the release notes.

I suggest we merge this PR as it is, release a new version, and do the "big" change in a separate PR. What do you think?

Hammster:
Great 🙂

I'm open to help if you need addition input on any further TypeScript definition support, maybe I'm able to create a proposal that will include all your requirements. 

Thank you for your time 👍

Simon-Laux:
Any updates on this?

Hammster:
I'm not sure as when this gets merged, it should work as-is I also updated the branch with the current revision of master.

y-lohse:
Sorry about the time it took to merge this one.

Hammster:
No problem :) I'm happy that it found it's way in.

PR #343, opened 2019-06-05
variableChangedEventCallbacks now preserved when VariableState is copied

michael-badrobotgames:
This missing line was breaking the variable observers.

### DIFF
--- a/src/VariablesState.ts
+++ b/src/VariablesState.ts
@@ -121,6 +121,7 @@ export class VariablesState{
 		this._defaultGlobalVariables = new Map(toCopy._defaultGlobalVariables);
 
 		this.variableChangedEvent = toCopy.variableChangedEvent;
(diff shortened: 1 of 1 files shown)


y-lohse:
Thank you! I had a quick look yesterday and the change seems good, but I'll have a more in depth look soon. Have you looked if the disabled unit tests are running again by any chance?

michael-badrobotgames:
Yes, with the fix the commented out tests now pass.

y-lohse:
That's really great news :)

I started the review but might get interrupted, so I'll start posting stuff here, sorry for the noise.
The reference implementation [doesn't have this line](https://github.com/inkle/ink/blob/0.8.3/ink-engine-runtime/VariablesState.cs#L120) so… that's odd. Maybe C# did the copy in some other way?

y-lohse:
Riiight, `variableChangedEventCallbacks` is inkjs specific. So the fix LGTM 👍

Issue #339, opened 2019-06-05
List creation fails on load in latest


joningold:
The following ink source:

LIST abc = A, B, C 

when exported and loaded via ink-js causes an error as the list definition is built. 

The exact error (in Chrome Version 74.0.3729.169) is 

Uncaught TypeError: Constructor Map requires 'new'

It's hard to unpick the TS / JS code, but it seems to happen in the constructor in ListDefinitionsOrigin.ts.

y-lohse:
Looking at it on discord, but so far I can't reproduce the issue doing this https://codesandbox.io/s/inklistdebug-4rl1d

y-lohse:
The problem happens when using `dist/ink.js`, ie. the legacy browser version. There must be a transpilation problem somewhere I guess.

ephread:
@y-lohse Ouch, I can investigate this over the weekend if you want.

y-lohse:
Let's say that the first one who takes a proper look at it posts here :)

ephread:
It's an issue with Babel. The first thing I'm going to attempt is to migrate to Babel 7, then, I'll report back!

ephread:
Missing configuration setting in the Rollup-Babel middleware, it should be:

```javascript
babel({
    exclude: 'node_modules/**',
    extensions: ['.js', '.ts'],
}),
```

I'll create a PR.

PR #346, opened 2019-06-07
fix: #339 - Babel transpilation not running correctly

ephread:
Move to Babel 7 and make sure Babel is run by Rollup.

You may want to double check that it _does_ solve #339 in the wild, as I only made sure the tests are green (they broke for `ink.js` before the fix).

By the way, we're not polyfilling Map and Set at the moment, would it make sense to do so for the legacy version?

Thanks for reviewing!


y-lohse:
Thanks, changes look good, I'll just do a local checkout later to check #339.

> By the way, we're not polyfilling Map and Set at the moment, would it make sense to do so for the legacy version?

Ehm yeah maybe. Support for them is really good, but still.
We should probably  [specify a target for babel](https://babeljs.io/docs/en/babel-preset-env#targets), something like `> 0.25%, not dead`. And active the `built-ins` option? [That looks ok](https://babeljs.io/repl/#?babili=false&browsers=%3E%200.25%25%2C%20not%20dead&build=&builtIns=usage&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDMBTA7jAsnADgCgEoBuAKCA&debug=true&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.4.5&externalPlugins=%40babel%2Fplugin-transform-object-assign%407.2.0).

PR #349, opened 2019-06-07
Switching to jest

y-lohse:
This PR switches from jasmine/istanbul to jest. @ephread would you mind having a look at this and tell me what you think?

The tests are run against the files in `engine/`, ie. the result of the typescript conversion, but before rollup does anything. It's probably better than running it against the rollup'ed file to get the correct coverage info, but it also means we can't test the compiled and legacy scripts in the same way.


.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": ["@babel/preset-env"],
+  "presets": ["@babel/preset-env", "@babel/preset-typescript"],
```
ephread: Is this required? Isn't this going to clash with `tsc`?

ephread:
> @ephread would you mind having a look at this and tell me what you think?

LGTM!

> It's probably better than running it against the rollup'ed file to get the correct coverage info, but it also means we can't test the compiled and legacy scripts in the same way.

Since the tests run fairly fast, couldn't we run them three times? The first time, we run them with coverage, against the transpiled runtime. The second time, we run them against the minified version and the third time, we run them against the legacy version?

I don't know how if we can achieve that by providing different config files to jest, but we can probably set an environment variable in NPM's `test` script and check the value of that variable in `tests/common.js`. Depending on its value, the script would either load `engine/runtime.js`, `ink-es2015.js` or `ink.js`.

Essentially, we'd have something like:

```json
{
  "scripts": {
    "test": "jest --coverage && INK_VERSION=minified jest && INK_VERSION=legacy jest"
  }
}
```

.babelrc
y-lohse: I thought it was required and produced no side effects, but… I just tried running the tests without it again and everything is fine.

y-lohse:
> Since the tests run fairly fast, shouldn't we run them three times?

Yep, I don't really have any other ideas to be honest. I quickly tried to load `ink-es2015.js` but jest can't parse the file and the error message doesn't show up, so we can give it a go in another PR.

ephread:
> I quickly tried to load ink-es2015.js but jest can't parse the file and the error message doesn't show up

There's something going on between terser and jest apparently, running the tests on the non-minified output for ink-es2015.js produces no issues. I'll investigate.

Issue #355, opened 2019-06-10
Version 0.9.0


y-lohse:
Change list here: https://github.com/inkle/ink/compare/0.8.3%E2%80%A60.9.0

Looks like there are a few big ones in there!

joethephish:
Just so you're aware: there's a big feature I added and then subsequently removed called "lookups". Added in 27f3cd7, with modifications added while I was working on other stuff, then eventually fully removed in b8e3547.

ephread:
Let’s get the party started! I’ll create a new branch and a new PR to track the migration (unless you get the chance to do it before me @y-lohse).

y-lohse:
Yep, no problem. Might be a while for me before I have time to do significant contributions, but I'll do my best!

y-lohse:
Yay this is published with v1.11.0 !

PR #363, opened 2019-06-14
[v0.9] Migration from 0.8.3

ephread:
Tackling #355.

Here's the list of changes:

- [x] CallStack **(MERGED)**
- [x] JsonSerialisation **(MERGED)**
- [x] StatePatch **(MERGED)**
- [x] Story **(MERGED)**
- [x] StoryState **(MERGED)**
- [x] VariablesState **(MERGED)**
- [x] SimpleJson **(MERGED)**

Feel free to:
- call dibs on specific files and create PRs against `v0.9`;
- discuss implementation details on this PR.

Remaining tasks:

- [x] Write unit tests for `SimpleJson.ts`
- [x] Use `SimpleJson.Reader` everywhere


y-lohse:
Started working on VariableState !

ephread:
Wow, it's been almost a year. Time flies! I'm currently porting the diff in inkgd, I'll resume work on inkjs soon after I'm done (probably next week or so).

One change isn't straightforward and it's in `SimpleJson.cs` (which doesn't have a counterpart). I believe JavaScript will convert `Infinity`, `-Infinity` and `NaN` into `null` when stringifying, but that doesn't match the new behaviour in upstream (`3.4E+38`, `-3.4E+38` and `0.0`). I don't know where's the best place to deal with this though.

What are your thoughts?

y-lohse:
Haven't really gotten back into this one either :/.

For the JSON things… if we do nothing, there would be a small incompatiblity, and a story exported with a NaN/Infinity probably couldn't be read again, since the value would be null. Not so great.
We could port SimpleJSON but the reason it's not there to begin with is that javascript has a good built-in JSON engine, so it's annoying to duplicate it.
Maybe we can look up the call chain — if it's only the FloatValues and IntValues that are converted like that, we may be able to override the `toJSON` function to match the behavior of upstream.

1. Port SimpleJSON
2. Ignore the specific
3.

ephread:
> Haven't really gotten back into this one either :/.

All good!

> We could port SimpleJSON but the reason it's not there to begin with is that javascript has a good built-in JSON engine, so it's annoying to duplicate it.

Yeah, I agree with not porting it unless we have no other options.

> Maybe we can look up the call chain — if it's only the FloatValues and IntValues that are converted like that, we may be able to override the toJSON function to match the behavior of upstream.

Oh yeah, that's smart, I didn't know it was possible. We have options then!

Now, about `JsonSerialisation.cs`. I guess we could match the changes, by wrapping a plain JavaScript object inside a `Writer` class, updating its inner hierarchy each time one of the `Write-` methods is called and finally, stringifying the complete object in`toString()` (since the stream-based methods are unlikely to be implemented in inkjs). It may also give us a place to convert `Infinity` & `NaN` if we can't bend `toJSON` to our will.

There's something I have trouble wrapping my head around however. [ink@0489e4db](https://github.com/inkle/ink/commit/0489e4dbd04d5157bba307441771a63acd29ee8c) describes the change as:

> Re-wrote JSON writing code to not create an intermediate Dict/List based representation, which created lots of garbage (and was a bit slower). Instead, we can now write to any stream.

but I'm not sure what _garbage_ refers to here.

@joethephish did you mean garbage in the final JSON file or during the transformation?

ephread:
I'm going to experiment a bit with `SimpleJson` and how we can port these changes. I'll report back soon!

joethephish:
Eeek I wouldn't want this silly NaN and Infinity thing to generate lots of work at your end, and definitely don't think an entire reimplementation of a JSON parser to be written. I'd rather remove the feature from the C# end. We only had one use case because we happened to use infinity as a special terminal value in some gameplay code that made its way into ink. But for the sake of not bloating inkjs it would be okay to revert that. Does the JSON spec not support infinity and NaN in general?

On garbage: I mean in the transformation. I was previously converting **Ink.Runtime native objects** --> **C# Dictionary/Lists** -> **JSON text**. But this generated an entire hierarchy of objects in the middle step that has to be garbage collected. Instead it now skips the middle man and serialises ink native objects directly to text.

ephread:
Thanks for jumping in @joethephish!

> Does the JSON spec not support infinity and NaN in general?

Nope, sadly. 😕

> Eeek I wouldn't want this silly NaN and Infinity thing to generate lots of work at your end, and definitely don't think an entire reimplementation of a JSON serialiser to be written. I'd rather remove the feature from the C# end. We only had one use case because we happened to use infinity as a special terminal value in some gameplay code that made its way into ink. But for the sake of not bloating inkjs it would be okay to revert that.

[`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter) can take a function as a parameter, which we believe, would allow us to easily replace `infinity` and `NaN` (I haven't tested it, but it looks very promising). I don't think this feature alone would bloat inkjs at all, it's the new serialiser that proves to be tricky to port (see below).

> On garbage: I mean in the transformation. I was previously converting Ink.Runtime native objects --> C# Dictionary/Lists -> JSON text. But this generated an entire hierarchy of objects in the middle step that has to be garbage collected. Instead it now skips the middle man and serialises ink native objects directly to text.

Ah, I see, thanks for the clarification! It makes sense. As I hinted above, the really complex part about 0.9.0 is `SimpleJson.Writer`. Instead of rewriting the serialiser, I'm thinking of having another intermediate representation of the state wrapped inside `SimpleJson.Writer` and then stringify everything at the end. That would require injecting a bit of custom code in `SimpleJson` (with its share of fresh bugs 😄), but it should be manageable.

In the end, I'm expecting the performances in JavaScript to remain fairly similar to what they were before, with the intermediate representation stored in `SimpleJson` instead of `Story`. The new upstream serialiser gives us a bit of work, but since it's performance-related on your end we can't expect you to remove it!

joethephish:
Just to check - you don’t think you can keep the code just the way it was before? (Other than whatever you want to do on NaN / Infinity)

Since it’s pure optimisation on the C# end and nothing has changed in the logic/format, maybe in this instance it would make sense to keep the JS version diverged, albeit with clear comments to demarcate where implementation varies?

ephread:
> Just to check - you don’t think you can keep the code just the way it was before? (Other than whatever you want to do on NaN / Infinity)

I feel a diverge would be harder to maintain in the long run, as we've tried to stick as close as possible to the C# source in the past. What do you think @y-lohse?

y-lohse:
> I feel a diverge would be harder to maintain in the long run, as we've tried to stick as close as possible to the C# source in the past. What do you think @y-lohse?

Yep I agree, I think it's ok to have small diversions but when it affects several hundred lines like in this case it's best to keep the 2 versions matching as closely as possible. Otherwise as you say future maintenance just gets harder.

That being said, I wouldn't mind releasing a new inkjs version without those changes as long as we keep an explicit task to port that file at a later time. Since it's been hard to find time to work on this release for a while now, we can be a bit more pragmatic 😛

joethephish:
Okay, no worries, whatever works for you guys!

ephread:
> Since it's been hard to find time to work on this release for a while now, we can be a bit more pragmatic 😛

Oh yeah, it's been tough.

I'm dedicating all my free time this week to inkjs 😄. I have quite a substantial amount of it due to COVID-19, so we should make significant progress!

y-lohse:
@ephread is there any more work to do on this PR? If not we can have a final review of this and merge it I believe :)

ephread:
@y-lohse nope, I'm just going to update the compatibility table you just approved but otherwise we're good to go!

ephread:
@y-lohse it's ready to be reviewed and merged. I pushed f564447 directly to v0.9 to avoid the back and forth of a PR, it's a tiny change.

I'll let you update the table in the bump commit!

ephread:
> Well… I think it's all good :) Great work 🎉

Thanks! I'll let you merge that one :)

y-lohse:
Cheers!

PR #367, opened 2019-06-17
Add coverage badge through Codecov

ephread:
[It works!](https://codecov.io/gh/y-lohse/inkjs) \o/

I'm remapping the coverage data with `remap-istanbul`. I also updated the badges in the README.

Thanks for reviewing!


ephread:
Ah, yeah, my bad. Didn’t see that it was deprecated.

So essentially, the lcov report generated by jest is not processed correctly. I didn’t really try hard to understand why, given Codecov gives no indication as to why the processing failed, but I suspect it’s because the paths inside the report are somewhat incorrect.

Let me see if I can find a way to fiddle with jest, Istanbul and remap without using the deprecated package!

ephread:
Upon further investigation, here's what's happening.

Jest doesn't pick up the TypeScript source maps, because it only does so when a transformer is involved. In our case, there are no transformers, since we're building the JavaScript files first, and then importing them in vanilla test cases. I see two options:

1. Keep things as they are and use `remap-istanbul` to manually remap the coverage.
2. Migrate the test cases to TypeScript; we could:
    - run them once through the TypeScript transformer and get coverage data;
    - then transpile them to JavaScript and run them against the two concatenated files.

I don't quite know how it would look like, but I believe it's possible.

y-lohse:
Ok, I see! I think ultimately 2 may be better… but I would say it's really low priority. The current setup does everything we need and using `remap-istanbul` is ok I think :)

ephread:
Alright, let's merge this PR and settle this when we'll update the test cases!

y-lohse:
Man that's a lot of dark orange for files that have >60% coverage 😁

ephread:
Haha, yeah, Codocov's pretty harsh! We could configure the threshold though. 😇

Issue #425, opened 2019-10-17
Impossible to set empty string as variable


RokasVaitkevicius:
Is it by design or is it a bug, that when I try to set variable to empty string it becomes number 0?
I found where this happens:
 https://github.com/y-lohse/inkjs/blob/0c6277ccb0b2493cc8b572e867f0da91d8f770e5/src/Value.ts#L23-L24

Basically this if returns true, because `Number('')` returns 0.

To solve this I would either move condition checking if type is string up the conditional chain, of course there might be a better solution to this.

Or am I just doing something wrong? If so, how to set variable to empty string using inkjs?

y-lohse:
No that looks like a bug. I think moving the string check up is a good solution, with a comment explaining why we need to diverge from the original implementation. Would you be able to make a PR for that?

RokasVaitkevicius:
Okay, will do that in a bit. As I understand inkjs follows C# lib where it is possible, even the sequence of conditions, so I should comment why it is diverged, right?

y-lohse:
Yes, that's correct! A link to this issue and a small text would be entirely enough :)

RokasVaitkevicius:
@y-lohse PR https://github.com/y-lohse/inkjs/pull/426 created. Let me know if I anything needs changing.

PR #426, opened 2019-10-17
Fix assigning variable to an empty string

RokasVaitkevicius:
Fixed https://github.com/y-lohse/inkjs/issues/425 issue


y-lohse:
Great, thank you!

PR #449, opened 2020-02-02
JObject changes

tcyrus:
* Stricter type checks (`===` instead of `==` where applicable)
* Use `object` while preserving compatibility with `JObject`

### DIFF
--- a/src/JObject.ts
+++ b/src/JObject.ts
@@ -1,1 +1,3 @@
-export type JObject = Record<string, any>;
+export interface JObject {
+	[key: string]: any;
+}
--- a/src/JObject.ts
+++ b/src/JObject.ts
@@ -1,3 +1,1 @@
-export interface JObject {
-	[key: string]: any;
-}
+export type JObject = Record<string, any>;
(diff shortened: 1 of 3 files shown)


y-lohse:
Hi @tcyrus and welcome!
The PR is a bit hard to review as it is, since it mixes a few different kinds of changes. Would you mind splitting the commits by intent, or even making one PR per intended change (formatting, equality checks, objects, etc)?

I'm happy to accept a PR for stricter equality. For `object` and `JObject` I'll need a more detailed look.
But I've also seen that some of the proposed changes (such as those in `JArrayToRuntimeObjList`) are syntax changes that we don't want in this repo, despite being common practice in js. This is because we try to keep the codebase as close to the C# base as possible, as it makes it easier to follow migrations. SO those sort of changes won't be accepted i'm afraid.

tcyrus:
Hey,

I realized that I made a lot more changes in my patch branch than I originally indented. I'm fine with closing this and splitting up my changes into more targeted PR's.

tcyrus:
@y-lohse Just out of curiosity, does this project have any style guidelines or reccomendations?

y-lohse:
I'm afraid not, I never got around setting any of them up. Feel free to do what works for you, as long as it doesn't cause major changes I'm good with it :)

PR #450, opened 2020-02-13
tests: Add test case for currentPathString

manuq:
Mainly to dispel a question asked in:
https://github.com/inkle/ink/issues/548


y-lohse:
Thank you!

PR #479, opened 2020-05-05
Fix #478

### LINKED ISSUE #478: NativeFunctionCall prints enum values in error messages, not value type strings
## Overview
I have noticed multiple times that runtime error messages produced by inkjs differ in slight but significant ways by those outputting by Inky. In particular, when one attempts to perform a logical operation like `>` on a variable which is not a number, the error message one should get looks approximately like this:

`Cannot perform operation > on String`

However, in inkjs, this looks like so:

`Cannot perform operation > on 3`

After a colleague pointed out this was likely referring to a type, not a value, we found that error messages from `NativeFunctionCall.ts` were including the value from the enum, which is an integer, rather than the key, which is a string.

## Reproduction
1. Produce a runtime error by comparing two variables which are of an incomparable type.
1. Note the error produced in inkjs differs from that produced by Inky.

## Pull request
https://github.com/y-lohse/inkjs/pull/479

furkle:
This should fix issue #478 by printing the keys of `ValueType` rather than the integer values.

### DIFF
--- a/src/NativeFunctionCall.ts
+++ b/src/NativeFunctionCall.ts
@@ -129,7 +129,8 @@ export class NativeFunctionCall extends InkObject{
 			if (this._operationFuncs === null) return throwNullException('NativeFunctionCall._operationFuncs');
 			let opForTypeObj = this._operationFuncs.get(valType);
 			if (!opForTypeObj) {
-				throw new StoryException('Cannot perform operation '+this.name+' on '+valType);
+        const key = Object.keys(ValueType)[valType];
+				throw new StoryException('Cannot perform operation '+this.name+' on '+key);
 			}
 
 			if (paramCount == 2) {
@@ -253,8 +254,10 @@ export class NativeFunctionCall extends InkObject{
 						parametersOut.push(castedValue);
 					} else
 						throw new StoryException('Could not find List item with the value ' + intVal + ' in ' + list.name);
-				} else
-					throw new StoryException('Cannot mix Lists and ' + val.valueType + ' values in this operation');
+				} else {
+          const key = Object.keys(ValueType)[val.valType];
+          throw new StoryException('Cannot mix Lists and ' + key + ' values in this operation');
+        }
 			}
 		}
 


src/NativeFunctionCall.ts
```diff
@@ -129,7 +129,8 @@ export class NativeFunctionCall extends InkObject{
 			if (this._operationFuncs === null) return throwNullException('NativeFunctionCall._operationFuncs');
 			let opForTypeObj = this._operationFuncs.get(valType);
 			if (!opForTypeObj) {
-				throw new StoryException('Cannot perform operation '+this.name+' on '+valType);
+        const key = Object.keys(ValueType)[valType];
```
ephread, line 132: Wouldn't `ValueType[valType]` work just as well? It's a bit more TypeScript-y, but I haven't tested it. 🙃

y-lohse:
Thanks for the bug report and fix, looks sensible to me as well! I'd prefer `ValueType[valType]`if it works as well, otherwise we can merge this as soon as the CI is satisfied :)

> By the way @y-lohse, we should probably create a PR template with a checklist mentioning the linter.

Yep, not the first time this happens 👍

ephread:
I run a contrived example with the current fix (`Object.keys`) and it still showed a number instead of the case name. Could it be possible that it worked on @furkle's setup but failed on mine? I assume it might be because `Object.keys` doesn't guarantee the same order on every runtimes, but I'm no JavaScript specialist. 

TypeScript enums are kinda like bidirectional maps and transpile to plain objects with both enum cases and enum values as keys. Perhaps the string keys are at the beginning of the array in some runtimes, while integer keys get the first spot on others?

Anyway, the good news is that I also tried with `ValueType[valType]` and it does the trick!

Oh and while I was working on `NativeFunctionCall` I noticed that [NativeFunctionCall.ts#L182](https://github.com/y-lohse/inkjs/blob/master/src/NativeFunctionCall.ts#L182) could be updated as well. 🙂

furkle:
@ephread @y-lohse Hm, interesting, not sure why `Object.keys` wouldn't have worked. It was a quick fix, though, so however you all want to fix it is fine, whether that's deleting the PR or just updating it 😃

ephread:
@furkle either way is fine, if you're happy to make theses changes, feel free to update the PR. Otherwise, let me know and I'll create a PR against your branch with the changes! 🙂

y-lohse:
Done in #505, thanks again!

PR #489, opened 2020-05-11
Port test suite from upstream

ephread:
## Description

It's something we discussed a bit a while ago, but with the move to 0.9.0 around the corner, I would feel more comfortable if we could have the test suite from upstream ported to inkjs. It would not replace what we currently have, but instead complement the current tests.

I went ahead and prototyped how it could look like, but to make code review manageable, I'm going to split the PoC in multiple PRs against this PR/branch. I'm reusing what I extracted for inkgd, so the port is fairly fast and straightforward. (With the few test cases I ported, coverage increased slightly and I discovered an difference in the way inkjs rounds numbers — maybe that was already known though).

I'd like to refactor the content of `tests/`:

- `files/`
    - `original/`: contains ink files separated in subdirectories (categories) (e. g. `builtins/`, `callstack/`, `choices/` — `test.ink` sits in an `inkjs/` subdirectory);
    - `compiled/`: contains the compiled json files and following the same structure as `original`;
- `specs/`
    - `ink/`: contains the new test cases ported from upstream, with one file per category;
    - `inkjs/`: contains the current test specs;
- `unit/`: (not sure about the name) but it would contain (if applicable) unit test created against the custom code we are writing for inkjs (e.g. `SimpleJson.js`);
- `common.js` would be slightly tweaked to support both test suites.
- `compile.js`: a small script used to compile the fixtures using inklecate (it's part of an npm task).

These changes are fairly minor, I'm sticking very close to what's already there.
I'm going to go ahead and create some PRs, so you get a better idea of what I'm suggesting.

Thanks for reading!

### Related PRs:

1. #490
2. #491
3. #492
4. #501

### Issues raised:

1. Can we use `-c` (count all visits) for all ink files? ([#491](https://github.com/y-lohse/inkjs/pull/491#discussion_r424367578))
2. Consider using [inklecate-node](https://github.com/furkleindustries/inklecate-node) to compile test files ([#491](https://github.com/y-lohse/inkjs/pull/491#discussion_r424371324))
3. Some tests are failing and have been disabled, they need to be investigated.

## Checklist

- [ ] The linter ran and found no issues (`npm run-script lint`).


y-lohse:
This all sounds quite good! I have a few general questions before diving into the code:

- The tests from upstream would have the same format than what we have now, eg. an ink.json file and a corresponding test case written in js?
- If I understand correctly, you want to split up the current `tests.ink` file that we have into multiple files? (Probably a good idea, just checking)
- I assume the files in specs/ink would mirror the files in files/original ? What about those on specs/inkjs ?
- Why do we rename `inkfiles` to `files`? I think it's good that the folder name explicitly says it contains ink files, whereas `files` seems very generic. `ink` could work too as a name I think.
- Nitpicking, but I'm not sure we need a separate `unit` folder? I like that the test suit really tests the output and not the implementation. We should totally have coverage for our custom code, but I think it could live with the other inkjs tests?


Apologies if some of these questions have obvious answers in the other PRs, I had a look at them but… well there's a lot going on 😅

ephread:
> This all sounds quite good!

Thanks! I'm going to address the crossed points as well.

> The tests from upstream would have the same format than what we have now, eg. an ink.json file and a corresponding test case written in js?

Yep! See #492 ([diff](https://github.com/ephread/inkjs/compare/test-compilation...ephread:test-cases)) for an example.

> ~~If I understand correctly, you want to split up the current tests.ink file that we have into multiple files? (Probably a good idea, just checking)~~

Not as part of this PR sequence. We could, but I'm not touching them for now since they work well. 

> ~~I assume the files in specs/ink would mirror the files in files/original ? What about those on specs/inkjs ?~~

That's correct, `specs/inkjs/*.spec.js` will use all use a single ink file stored in `file/[original|compiled]/inkjs`. It's a bit weird, but it makes the compile script easier to write 😄.

> Why do we rename inkfiles to files? I think it's good that the folder name explicitly says it contains ink files, whereas files seems very generic. ink could work too as a name I think.

Because I moved stuff around and reused code 😅, but there's no reason to name it `files` over `inkfiles` or `ink`. That's a good point, I'll rename the directory!

> This is already a big step, so I'm not saying we need to do the slicing up of the existing tests as part of this PR, but do you think it would be a problem in the future? Would we have name collisions between the ink and inkjs tests?

Well, as long as they are named differently or sit in their specific subdirectories, it should cause no problem at all.

Good point! It absolutely could live with the other inkjs tests. I'd still like to write a couple of unit tests for inkjs's custom code (such as the new `SimpleJson.ts`), to make sure the implementation is correct (not too confident in my ability to write bugless code).

> but do you think it would be a problem [to slice the ink file up] in the future? Would we have name collisions between the ink and inkjs tests?

y-lohse:
> Well, as long as they are named differently or sit in their specific subdirectories, it should cause no problem at all.

I'm just a bit concerned that we'd want something like `files/original/choices.ink` twice. But I guess we can figure it out when we split the current test file.

All good for the other points 👍 !

ephread:
Ready to merge when approved. I'll open new issues to address the points raised during review, we can do this after 0.9.0.

ephread:
Oh yeah! 😎

PR #490, opened 2020-05-11
[Tests 1/4] chore(tests): update project structure to support upstream test suite

ephread:
## Checklist

- [x] The new code additions passed the tests (`npm test`).
- [x] The linter ran and found no issues (`npm run-script lint`).

## Description

Following #489.

Sorry about the noise, the editor removed trailing spaces. Changes are:

- `common.js` was updated to support the `category` parameter;
- specs were moved into the `inkjs` subdirectory;
- `tests.ink` and `tests.json` were respectively moved to `files/original/inkjs` and `files/compiled/inkjs`; note that `tests.json` was renamed to `tests.ink.json`, it's pure convenience, since the compilation script (I'll add it in another PR) does not strip `.ink` from the output.

Let me know what you think, even on petty changes like `Strip the BOM.` 😁

PR #491, opened 2020-05-11
[Tests 2/4] chore(tests): add compile scripts and ink files

ephread:
## Checklist

## Description
Depends on #490, so I'll left it as a draft until it's merged. Here is the [**proper diff**](https://github.com/ephread/inkjs/compare/test-structure...ephread:test-compilation).

Don't be scared by the number of files changed, these are mostly ink files and their JSON representation. The two important files are:

- `package.json` where I added a new dev dependency, `fs-extra`.
- `compile.json`: which wraps inklecate and compile all the files (it can be useful to recompile to test between different versions of inklecate).


tests/compile.js
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env node
+
+// Recompile baseline ink files with the current version of inklecate avilable
```
y-lohse, line 3:
```suggestion
// Recompile baseline ink files with the current version of inklecate available
```

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+// $PATH.
```
y-lohse, line 4: hu? Is that part of the previous comment?

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+
+let childProcess = require('child_process');
+let fs = require('fs-extra');
+let path = require('path');
+
+let fileDirectory = path.join(__dirname, 'files');
+let inkFileDirectory = path.join(fileDirectory, 'original');
+let compiledFileDirectory = path.join(fileDirectory, 'compiled');
+
+// These files require the `-c` flag so that all visits will
```
y-lohse, line 14: Could we use that on all files?

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+// be counted.
+let filesRequiringCFlag = [
+	'visit_counts_when_choosing.ink',
+	'turns_since_with_variable_target.ink',
+	'read_count_variable_target.ink',
+	'tests.ink'
+]
+
+function getAllInkFiles(dir) {
```
y-lohse, line 23: Ok for me but if we want less code here, something like [`glob`](https://www.npmjs.com/package/glob) would probably work just as well :)

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+				return files.concat(path.join(dir, file));
+			} else {
+				return files;
+			}
+		}
+	}, []);
+}
+
+function runInklecate(input, output, extraArgs) {
+	let command = `inklecate ${extraArgs} -o "${output}" "${input}"`
```
y-lohse, line 44: I wonder if we should use something like https://github.com/furkleindustries/inklecate-node instead of expecting `inklecate` to be in PATH? Or potentially include a version of inklecate in this repo if we want tight control over what version we use?

I personally need an ink compiler on a semi regular basis when working or debuging things in inkjs and usually end up downloading the latest version again. Having a compiler ready would come in handy.

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+				reject(error);
+			} else {
+				resolve(stdout);
+			}
+		});
+	});
+}
+
+function compileInkFile() {
+	console.log("Compiling test cases…");
```
y-lohse, line 58: ... -> 🙄
… -> 😍

tests/compile.js
```diff
@@ -0,0 +1,91 @@
...
+				return runInklecate(file, out, extraArgs);
+			} else {
+				return fs.ensureDir(outDirectory).then(() => runInklecate(file, out, extraArgs));
+			}
+		})
+	})
+
+	Promise
+		.allSettled(promises)
+		.then(
```
y-lohse, line 85: Note that we're on node 12 now, we could use `async/await` :)

tests/compile.js
ephread: Cheers!

tests/compile.js
ephread: Er, yeah… ? Haha, yes it's confusing, but it might go away, see other comment.

tests/compile.js
ephread: I don't really know, it may be something we need to ask Joe about. I sticked to what was done in the `Test.cs` and not all files are compiled with all visit counts enabled.

tests/compile.js
ephread: Absolutely. I didn't know if you were okay adding a package like glob. I'll happily rewrite this!

y-lohse: As you want, I don't mind it as it is :)

Generally speaking I have 0 issues adding packages as dev dependencies. If they get shipped with inkjs it's a different story :)

tests/compile.js
ephread: I took a very hands-off approach, putting the burden of configuring inklecate on the user. I maintain [homebrew taps for macOS](https://github.com/ephread/homebrew-ink) and I use the versioning capabilities to switch between different versions of inklecate (I would love if someone could do the same for linux and windows).

But since this is a JavaScript project, you're right, it makes a lot of sense to leverage the excellent work from furkle. The only downside I see is that we have no ways of switching between different inklecate versions or know which one is currently bundled. Maybe that could be fixed with a PR?

I think bundling `inklecate` on our own would be overkill though.

I'll have a deeper look at these options at the very end!

tests/compile.js
ephread: Oh I'm terrible at Javascript man! How would you write it?

tests/compile.js
ephread: Haha, you need to know it's actually doing something, don't you? 😄

tests/compile.js
y-lohse: > The only downside I see is that we have no ways of switching between different inklecate versions or know which one is currently bundled. Maybe that could be fixed with a PR?

Yeah this is my concern as well. i don't think we'd ever want to go back to older inklecate versions, but we'd probably want it to be relatively up to date.

@furkle Do you keep inkleate up to date on `inklecate-node`? Or would you accept PRs that do it?

> I'll have a deeper look at these options at the very end!

Yeah this isn't blocking at all, no problem.

ephread: > i don't think we'd ever want to go back to older inklecate versions

Actually, I'm fairly confident furkle keeps it up to date. But, for inkgd, I juggled between 0.8.2, 0.8.3 and 0.9.0. I wanted to make sure that the current runtime worked with older versions of inklecate as well (as long as they shared the same inkVersion internally), that's what I meant :smile:

Might happen at some point with inkjs as well! Oh an maybe it'd be a good idea to have a matrix in the CI and test with multiple different versions of inklecate?

tests/compile.js
y-lohse: Fair warning, I didn't actually test this 😁:

```
async function compileInkFile() {
  console.log("Compiling test cases…");

  let files = getAllInkFiles(inkFileDirectory);

  let promises = files.map(async file => {
    let out = path.join(
      compiledFileDirectory,
      path.relative(inkFileDirectory, file) + ".json"
    );
    let fileName = path.basename(file);

    let extraArgs = "";

    if (filesRequiringCFlag.includes(fileName)) {
      extraArgs = "-c";
    }

    let outDirectory = path.dirname(out);

    const exists = await fs.pathExists(outDirectory);

    if (exists) {
      return runInklecate(file, out, extraArgs);
    } else {
      await fs.ensureDir(outDirectory);
      return runInklecate(file, out, extraArgs);
    }
  });

  let results = await Promise.allSettled(promises);
  let errors = results.filter(result => result.status === "rejected");

  if (errors.length === 0) console.log("Done.");
  else errors.forEach(error => console.log(error.reason.message));
}
```

Also instead of 

```
const exists = await fs.pathExists(outDirectory);

    if (exists) {
      return runInklecate(file, out, extraArgs);
    } else {
      await fs.ensureDir(outDirectory);
      return runInklecate(file, out, extraArgs);
    }
```

… I think we could do only:

```
await fs.ensureDir(outDirectory);
return runInklecate(file, out, extraArgs);
```

ephread: Well, you're good at this, you should consider make a career out of it! 😄
It's working very, very well, I only had to make the map callback async.

Thanks for the help!

tests/compile.js
y-lohse: >  I wanted to make sure that the current runtime worked with older versions of inklecate as well

You're right, that makes sense. A matrix in the CI would make sense to test several inklecates, but now I'm not sure how we'd have several versions to begin with. Except if we keep inklecate versions in the repo, but as you say it might be overkill.

ephread:
The last push should fix most problems, I'll keep track of pending issues in #489.

tests/compile.js
ephread: We can open an issue about this after 0.9.0!

y-lohse:
Yep this looks good to merge I think !

PR #492, opened 2020-05-11
[Tests 3/4] tests: add builtin test cases

ephread:
## Checklist

- [x] The new code additions passed the tests (`npm test`). **Almost!**

## Description

Depends on #490 and #491. Here is the [**proper diff**](https://github.com/ephread/inkjs/compare/test-compilation...ephread:test-cases).

I added a test file for the `builtin` category. Two things to note:
- I named the file Builtins.**test**.js instead Builtins.**spec**.js and used `test` instead of `it`, since tests in upstream are not written in a spec-style;
- I disabled the first test ("floor, ceiling and casts") because it's failing — once all the tests are ported, I'll investigate why some of them are failing.


y-lohse:
> I named the file Builtins.test.js instead Builtins.spec.js and used test instead of it, since tests in upstream are not written in a spec-style;

It makes sense, but it does irk me a bit to have 2 different formats. And the folder is called `specs` too 😁. My concern is that when writing tests, we'd have to check the folder we're in to decide to use `it` or `test` or `spec` etc. Does that make sense? I don't have a strong opinion on which one is better between the two styles though.

> I disabled the first test ("floor, ceiling and casts") because it's failing — once all the tests are ported, I'll investigate why some of them are failing.

No problem!

_____

Had a quick look at this diff, looks good so far :)

ephread:
> It makes sense, but it does irk me a bit to have 2 different formats. And the folder is called specs too 😁. My concern is that when writing tests, we'd have to check the folder we're in to decide to use it or test or spec etc. Does that make sense? I don't have a strong opinion on which one is better between the two styles though.

Haha, true the directory is named `specs`! It's a bit confusing. We can use `specs` and `it` for everything, it just that I'll write `it('tests blablabla')` instead of `test('blablabla')`. If you're fine with it, I'll replace them all!

y-lohse:
>  If you're fine with it, I'll replace them all!

Yep that's totally ok fo me !

PR #495, opened 2020-05-12
[v0.9] Callstack.ts

ephread:
## Checklist

- [ ] The new code additions passed the tests (`npm test`). **N/A**

## Description
Thanks for taking the time to review!

### DIFF
--- a/src/CallStack.ts
+++ b/src/CallStack.ts
@@ -365,7 +365,6 @@ export namespace CallStack {
 		public WriteJson(writer: SimpleJson.Writer) {
 			writer.WriteObjectStart();
 
-			// callstack
 			writer.WritePropertyStart('callstack');
 			writer.WriteArrayStart();
 			for (let el of this.callstack) {
@@ -390,7 +389,6 @@ export namespace CallStack {
 			writer.WriteArrayEnd();
 			writer.WritePropertyEnd();
 
-			// threadIndex
 			writer.WriteIntProperty('threadIndex', this.threadIndex);
 
 			if (!this.previousPointer.isNull) {
--- a/src/CallStack.ts
+++ b/src/CallStack.ts
@@ -11,6 +11,7 @@ import {Container} from './Container';
 import {Debug} from './Debug';
 import {tryGetValueFromMap} from './TryGetResult';
 import {throwNullException} from './NullException';
+import {SimpleJson} from './SimpleJson';
 
 export class CallStack{
 	get elements(){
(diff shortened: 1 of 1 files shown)


src/CallStack.ts
```diff
@@ -353,33 +362,44 @@ export namespace CallStack {
...
+			writer.WritePropertyEnd();
 
-			threadJObj['callstack'] = jThreadCallstack;
-			threadJObj['threadIndex'] = this.threadIndex;
+			// threadIndex
+			writer.WriteIntProperty('threadIndex', this.threadIndex);
 
 			if (!this.previousPointer.isNull) {
 				let resolvedPointer = this.previousPointer.Resolve();
 				if (resolvedPointer === null) { return throwNullException('this.previousPointer.Resolve()'); }
```
y-lohse, line 398: Probably a silly question but I'm a bit rusty, remind me why we have these extra lines? Is it to please the typescript god?

src/CallStack.ts
```diff
@@ -353,33 +362,44 @@ export namespace CallStack {
 			return copy;
 		}
 
-		get jsonToken(){
-			let threadJObj: any = {};
+		public WriteJson(writer: SimpleJson.Writer) {
+			writer.WriteObjectStart();
 
-			let jThreadCallstack: any[] = [];
+			// callstack
```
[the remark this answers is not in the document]
ephread: Yup, I either copy pasted the old code that was ported before that rule was fully enforced or copy pasted from `Callstack.cs`. I'll remove them!

src/CallStack.ts
ephread: No silly questions, I actually had to be reminded myself.

It's to please the typescript god indeed and follow the behaviour of C#. Typescript will complain if we try to access a variable which might be null, but is smart enough to detect paths where it can't be null since a check was performed. We have these lines in many of placed, they'll raise a "NullPointerException" like C#.

As I got back into it, I took notes of all the strange things. We really need to work on this  CONTRIBUTING document!

y-lohse: What I find strange is that there wasn't a reason to use `previousPointer` or `resolvedPointer` previously, so… why was it a problem if they were null ?
Anyway, they make sense now!

ephread: > What I find strange is that there wasn't a reason to use previousPointer or resolvedPointer previously, so… why was it a problem if they were null ?

I'm not following, what do you mean?

y-lohse: Nevermind, I misread something in the C# diff which threw me off. It all makes sense now 👍

PR #496, opened 2020-05-12
[v0.9] Story.ts

ephread:
## Checklist

- [ ] The new code additions passed the tests (`npm test`). **N/A**

## Description
Thank you for having a look!

### DIFF
--- a/src/Story.ts
+++ b/src/Story.ts
@@ -172,8 +172,13 @@ export class Story extends InkObject{
 		// ------
 	}
 
-	public ToJson(writer?: SimpleJson.Writer) {
+	// Merge together `public string ToJson()` and `void ToJson(SimpleJson.Writer writer)`.
+	// Will only return a value if writer was not provided.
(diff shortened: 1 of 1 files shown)


src/Story.ts
```diff
@@ -171,17 +172,42 @@ export class Story extends InkObject{
...
+			writer.WriteObjectEnd();
+			writer.WritePropertyEnd();
+		}
 
-		if (this._listDefinitions != null)
-			rootObject['listDefs'] = JsonSerialisation.ListDefinitionsToJToken(this._listDefinitions);
+		writer.WriteObjectEnd();
 
-		return JSON.stringify(rootObject);
+		return writer.ToString();
```
y-lohse, line 210: Since we don't quite mirror the C# version here (we always return a string, the C# version has a definition that has a void return), maybe we could add a comment to make it explicit?

ephread:
Let me know what you think of the fix @y-lohse

y-lohse:
Yep sorry, saw it and didn't comment. LGTM 👍

PR #501, opened 2020-05-13
[Tests 4/4] All test cases

ephread:
## Checklist

- [x] The new code additions passed the tests (`npm test`). **Almost!**

## Description

Depends on #490, #491 and #492. Here is the [**proper diff**](https://github.com/ephread/inkjs/compare/test-cases...ephread:test-more-cases).

I disabled a couple of tests again, I'll investigate them later.


ephread:
> That's… a lot of tests :D I hadn't realized how much work went into it, awesome job.

Thank you! Most of it was already done, it only took a few hours to convert the test cases 🙂

> It's a bit too much to review in detail, but it all looks good. We use var story; instead of let story; but it's a detail, we can always change that at some other point.

I turned all `var`s into `let`s in all the test files. Tests are all green, so I'll merge the PR later today!

PR #502, opened 2020-05-13
[v0.9] VariablesState.ts

ephread:
## Checklist

## Description

I completed what @y-lohse started (and practically finished at the time). Three things to take note about:

1. I replaced `== null` by `=== null`, I don't think it makes much difference, but I think we have strict equality almost everywhere else. I hope that's okay!
2. I added `isEquatable` in `TypeAssertion.ts`, so that `RuntimeObjectsEqual` can compare `Inklist` and `Path` using `Equals` while using `===` for anything else.
3. I moved `SetJsonToken` and `ApplyStatePatch` up, so they match the position of their C# counterparts. Was there a reason for them to be at the bottom?

Also, we have three different ways of converting `TryGet` calls 😂. We really need to settle on one syntax!

### DIFF
--- a/src/TypeAssertion.ts
+++ b/src/TypeAssertion.ts
@@ -45,6 +45,10 @@ export function nullIfUndefined<T>(obj: T | undefined): T | null {
 	return obj;
 }
 
+export function isEquatable(type: any) {
+	return (typeof type === 'object' && type.Equals === 'function');
+}
+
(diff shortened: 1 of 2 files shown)


y-lohse:
I think the double equals in the position of the new functions were things I left to do later on, not thinking it'd be so long 😔. Totally the right call to change that.

The `isEquatable` is a clever solution, I like it!

PR #505, opened 2020-05-15
Fix Value type errors

y-lohse:
I'd like to publish of minor version of inkjs before rolling out the changes from [v0.9](https://github.com/y-lohse/inkjs/pull/363) which are more substantial, so this is #479 with the changed that had been discussed. 

@furkle can you confirm that this PR fixes the original issue?

### DIFF
--- a/src/NativeFunctionCall.ts
+++ b/src/NativeFunctionCall.ts
@@ -129,7 +129,7 @@ export class NativeFunctionCall extends InkObject{
 			if (this._operationFuncs === null) return throwNullException('NativeFunctionCall._operationFuncs');
 			let opForTypeObj = this._operationFuncs.get(valType);
 			if (!opForTypeObj) {
-        const key = Object.keys(ValueType)[valType];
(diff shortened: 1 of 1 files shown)


src/NativeFunctionCall.ts
```diff
@@ -129,7 +129,8 @@ export class NativeFunctionCall extends InkObject{
 			if (this._operationFuncs === null) return throwNullException('NativeFunctionCall._operationFuncs');
 			let opForTypeObj = this._operationFuncs.get(valType);
 			if (!opForTypeObj) {
-				throw new StoryException('Cannot perform operation '+this.name+' on '+valType);
+		const key = ValueType[valType];
```
ephread, line 132: Its curious the linter isn't complaining about the indentation.

furkle:
I'm having some issues reproing this on my end, which I don't believe relate to the code changes, so probably good to go ahead and push it. Thanks again!

PR #519, opened 2020-05-25
chore: Switching to eslint

y-lohse:
## Checklist

## Description

[tslint is deprecated](https://github.com/palantir/tslint) which is cool because it was acting up for me anyway. This PR switches the project to eslint.

I turned off a lot of rules that probably make sense in the grand scheme of things but would require changes to the codebase. The ones I left in are those that can be automatically fixed.

Since this PR will change a lot of files and we have #363 going on, I'll continue working on this after we're done.

- [x] Run the automatic fix
- [x] Convert the `tslint:disable` comments to `eslint-disable-whetever`
- [x] Check the warnings, either fix the problems or disable them if they make sense for C# compat
- [x] Final review of the eslint config file


ephread:
Yay! That's a great news, the current config ignores a lot of things, like spaces between braces, brackets, etc. It's super inconsistent, if we can pick a style and fix all of them that would be awesome!

y-lohse:
In the end we have 17 disabled rules from the recommended version, some of them only trigger warnings, some only a handful of errors, and some produce several hundred errors. I also don't think a lot of indentation rules are enforced through the recommended config, so that's another aspect we can improve upon in another PR.

src/engine/Pointer.ts
```diff
@@ -6,55 +6,55 @@ export class Pointer{
...
-	}
-
-	public static StartOf(container: Container | null): Pointer{
-		return new Pointer(container, 0);
-	}
-
-	public static get Null(): Pointer {
-		return new Pointer(null, -1);
-	}
+  constructor();
```
ephread, line 9: There are spaces here, then tabs. That's weird, shouldn't the linter complain? But I guess that's what you meant by:

> I also don't think a lot of indentation rules are enforced through the recommended config

y-lohse: Yeah surprisingly there are no default tules for this.

y-lohse:
I'm going to merge this as it is and introduce a second PR with only spacing changes because… it's going to be a LOT of changes.

y-lohse:
Fiiinally.

Issue #523, opened 2020-05-25
Run the test suites against both ink(-es2015).js and the TypeScript sources


ephread:
After v0.9 is merged I'd like to make a few changes to our test architecture (again).

I want to run the tests both with TypeScript and JavaScript! Three reasons:

1. I can't reasonably debug the JavaScript tests, since they rely on the transpiled/rolluped code; I have to use stuff like `console.log`, it slow, clunky and it takes forever to rebuild;
2. the coverage metrics of the tests written in TypeScript are not included the final report;
3. if we run the tests in TypeScript directly, we should be able to get rid of the deprecated `remap-istanbul` I put in a while ago and include all the tests in the coverage.

To summarise: TypeScript for debug & coverage, JavaScript to make sure the babelised-rolluped code is still passing the tests.

I plan to move the tests in TS (it's very quick), but also compile them to JS. By using the environment variables, I'll be able to choose which `Story` to load in `common.ts`. When jest is ran against the JavaScript files, it'll load `ink.js` and `ink-es2015.js`. When it's ran against TypeScript file, it'll load the TypeScript class directly.

Something like:

```json
{
    "test": "npm run-script test:javascript && npm run-script test:typescript",
    "test:javascript": "INK_TEST=dist jest --config=jest.javascript.config.js && INK_TEST=legacy jest --config=jest.javascript.config.js",
    "test:typescript": "jest --coverage",
}
```

I'll need to work out the type definition (probably just a `declare module "*";` somewhere) for the compiled files, but it should be doable!

What do you think?

y-lohse:
I understand the motivation and mostly how we'd do it.

> I plan to move the tests in TS (it's very quick), 

Changing the extension and slapping a few `:any`, or did you have more in mind?

> but also compile them to JS. 

When would we do this? Still on the CI as part of the testing step?

_______

My other slight concern is execution time, the Ci task takes more than a minute now. We should be able to use travis build stages and run all 3 test suites at the same time to speed things up. Probably best to do this in a separate PR though.

ephread:
> Changing the extension and slapping a few :any, or did you have more in mind?

Nope, that's about it!

> When would we do this? Still on the CI as part of the testing step?

Good question, probably as part of the step test. Compiling the specs should be very fast.

> My other slight concern is execution time, the Ci task takes more than a minute now.

[...]

PR #526, opened 2020-05-26
Typescript tests

ephread:
## Checklist

## Description

Implements #523. It's a big PR, sorry. It's going to be annoying to review. It can either be included in v0.9 (the branch I created against) or merged into master later on.

I'll guide you through the points of interest.

1. I moved `tests` under `src` so it can be compiled by TypeScript.

2. [`.npmignore`](https://github.com/y-lohse/inkjs/pull/526/files#diff-0fd4ef892d9d4990033701887c2f9bcc) – while working on the PR, I noticed we were including a lot of unnecessary things in the npm package. I updated the file to remove the clutter.

3. `jest` configs, the [default config](https://github.com/y-lohse/inkjs/pull/526/files#diff-2d0cd5d10b9604941c38c6aac608178a) file is now TypeScript-only and generates coverage. I added a second configuration file ([`jest.config.javascript.js`](https://github.com/y-lohse/inkjs/pull/526/files#diff-68617a8b7f2a203b6646cc617c2c8593)) for JavaScript.

4. I revamped the scripts in [`package.json`](https://github.com/y-lohse/inkjs/pull/526/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2). Now we have:

```json
{
    "test": "test:typescript && test:javascript",
    "test:typescript": "jest",
    "test:javascript": "test:javascript:dist && test:javascript:legacy",
    "test:javascript:dist": "jest",
    "test:javascript:legacy": "jest",
}
```

5. I removed all unused destructured variables that lead to collisions with terser (generalisation of #467). If I can reproduce the problem with a minimal example, I'll submit an issue.

6. I tweaked [common.ts](https://github.com/y-lohse/inkjs/pull/526/files#diff-7037773fb8cac60e88e7a63ba1a6e150) so that it returns a _different_ `Story` depending on the environment variables. It's either inkjs or the imported `Story`.

7. I slightly tweaked the _Exported classes_ spec in [Integration.spec.ts](https://github.com/y-lohse/inkjs/blob/c5c7a274bf0a208504e48e77e96f9937a27d6168/src/tests/specs/inkjs/Integration.spec.ts#L242). It's only executed in JavaScript.

Thanks for reviewing, now I can debug the tests!

**==== Edit ====**
With the new debugging capabilities, I was able to fix all the disabled tests! But there's a catch. I'll create a separate PR once this one is merged so we can talk about it.


y-lohse:
Thanks for the walkthrough, that should help! It looks like 937aeb779a58cc310aed6d290fa36100d6a7075e is part of this PR but it shouldn't?

I'll review the rest of the PR in the next couple of days, though ideally we'd merge v0.9 before.

ephread:
> Thanks for the walkthrough, that should help! It looks like 937aeb7 is part of this PR but it shouldn't?

You're welcome, that's weird I'll have a look once v0.9 is merged!

package.json
```diff
@@ -5,11 +5,11 @@
...
   "types": "ink.d.ts",
   "scripts": {
-    "test": "INK_TEST=dist jest && INK_TEST=legacy jest && jest --coverage && npm run-script test:remap",
-    "test:remap": "npm run-script test:remap:lcov && npm run-script test:remap:html && npm run-script test:remap:json",
-    "test:remap:lcov": "remap-istanbul --input coverage/coverage-final.json --output coverage/lcov.info --type lcovonly",
-    "test:remap:html": "remap-istanbul --input coverage/coverage-final.json --output coverage/lcov-report --type html",
-    "test:remap:json": "remap-istanbul --input coverage/coverage-final.json --output coverage/coverage-final.json",
+    "test": "npm run-script test:typescript && npm run-script test:javascript",
+    "test:typescript": "jest",
+    "test:javascript": "npm run-script build && npm run-script test:javascript:dist && npm run-script test:javascript:legacy",
```
y-lohse, line 10: I don't think we should add the `build` step here, otherwise we'll run thebuild multiple times on the CI.

ephread:
> Would it be possible/practical to leave the typescript tests in tests after all? Something along the lines of adding a dir in the includes section of tsconfig?

Sadly, as far as I know, no. I tried to have them outside of the `rootDir`, but TypeScript wasn't able to resolve them. It might be possible as the cost of an extremely convoluted set of `tsconfig`s, but I'm not sure.

We only have `inkfiles` and `compile.js` in `tests` though. Maybe we should rename the directory to something else, perhaps `cases` or `fixtures`? We can also move them to `src/tests`.

package.json
ephread: On multiple occasions, I forgot to rebuild before running the tests, asking myself why my changes did nothing. For my own sanity, I added the build phase as a dependency of the test phase. But since JavaScript tests won't be used for debugging in the future, I will no longer pull my hair out.

I'm happy to revert that change!

y-lohse:
> We only have inkfiles and compile.js in tests though. Maybe we should rename the directory to something else, perhaps cases or fixtures? We can also move them to src/tests.

Ah that's right. IMO moving them to src/tests is a good option, so all test-related files are in the same place. Which was my main concern anyway 😁

package.json
y-lohse: I generally start a build-watch in a separate console to avoid that, but I get the pain :)

ephread:
@y-lohse I moved the files to `src/tests` 🙂

Issue #147, opened 2018-06-05
Add test for list shuffling


y-lohse:
After #145, it'd be good to have a test case covering that. Testing randomness can't be difficult, can it?

ephread:
@y-lohse I believe this one can be closed. There's now tests covering shuffles ([here](https://github.com/y-lohse/inkjs/blob/8e4f2a2e6ca4f3bff7c4c6c3bb59728164590a43/src/tests/specs/ink/Sequence.spec.ts#L38) and [there](https://github.com/y-lohse/inkjs/blob/8e4f2a2e6ca4f3bff7c4c6c3bb59728164590a43/src/tests/specs/ink/Sequence.spec.ts#L46) for instance).

y-lohse:
👍 thanks!

PR #541, opened 2020-06-05
fix: disabled tests

### LINKED ISSUE #409: Incorrect float operations
I've run into a strange bug, and the more I look at it, the less sure I am of what exactly is going wrong. Consider the following ink script:
```
# title: bug

~ SEED_RANDOM(1234)
~ temp r = RANDOM(1, 99)
~ temp v = FLOAT(r) / FLOAT(100)
"r" is: { r }
"v" is: { v }
{ v > 0 && v < 1: "v" is within expected bounds. }
```

This is simple:
1. We fix the random number generator for consistent results between runs on any given ~platform~ ink runtime.
2. We generate a random number and divide by `FLOAT(100)`, to ensure we get a fractional result, at least on C# (I don't think #39 applies here - as we'll see in a minute).
3. We print the results and an optional notice if we got an expected result. By definition, we always should get this notice: the lowest value possible is `0.01`, and the highest is `0.99`.

Running via `inklecate -p` yields the expected results:
```
❮ inklecate -cp bug.ink
 ﻿﻿ "r" is: 44
# tags: title: bug
"v" is: 0.44
"v" is within expected bounds.
```

However, running via inkjs, we get something quite unexpected:
```
> .ink start bug


[bot]
"r" is: 32
"v" is: 0

>


[bot]
📕 Ended current story 📕
```

Frankly, I'm a little mystified. I'm aware of some of the parsing difficulties with floats that are actually whole numbers (as talked about in #39). But my testing in a node repl indicates that we should never get 0 as a result, no matter what:
```javascript
❯ node
Welcome to Node.js v12.8.0.
Type ".help" for more information.
> 32 / 100
0.32
> 32 / 100.0
0.32
> parseInt(32) / parseInt(100)
0.32
> parseInt(32) / parseInt(100.0)
0.32
>
```

What seems to be happening is that we're coercing the _result_ of the operations to an integer at some point:
```javascript
❯ node
Welcome to Node.js v12.8.0.
Type ".help" for more information.
> parseInt(0.32)
0
>
```

I've tried to follow the engine code, looking for where this might happen. The only place I can find where we call `parseInt` is when casting a `StringValue` to an `IntValue`. But I can't see where or why we'd need to cast _from_ a string _to_ an int with this simplistic ink script. Why would we? The operations are all numeric, up until the point where we print them out - and even then, we should be going from some subclass of `Value<number>` to `StringValue`, not the other way around.

Any ideas? Do we unilaterally coerce numeric results to `IntValue` rather than detecting the type of the result? That's the only realistic thing I can think of, but I can't quite follow the Value code well enough to see if that's the case.

ephread:
## Checklist

## Description

This PR re-enable all disabled tests:

- there's a new test case for 0.9.0. The expected output is slightly different, because of the PRNG. I believe #147 can be closed.
- for arithmetics tests, I either changed the expected output to match inkjs's internal behaviour or used a workaround.

I'll recap what's going on, so the problem I'm trying to solve is easier to grasp.

------

The nature of `number`s in JavaScript leads to two major differences with upstream.

1. JSON numbers are deserialised as floats, regardless of how they are written (`1.0` vs. `1`). They are later boxed in either an `IntValue` or a `FloatValue` depending on the result of `Number.isInteger()`. This is in contrast with upstream, where `1.0` will become a `FloatValue` and `1` will become an `IntValue`. In practice, this means that loading the state/story will erase the type for ambiguous values. We can't solve this issue unless we roll out a custom parser, I plan of explaining this limitation in the README.

2. Another difference with upstream is arithmetic. There's no true integer arithmetic in inkjs, which leads to subtle differences. [Flooring the result of integer division](https://github.com/ephread/inkjs/blob/e4eab7c1ab1cad69a82c258007142a425804b5c7/src/engine/NativeFunctionCall.ts#L328) or the result of [INT()](https://github.com/ephread/inkjs/blob/e4eab7c1ab1cad69a82c258007142a425804b5c7/src/engine/NativeFunctionCall.ts#L377) ensures that the return type will be boxed again as an `IntValue`. But with `FLOOR()` and `CEILING()`, it's undesirable to get an `IntValue` from a `FloatValue`. Take the following example:

```
{CEILING(1.2) / 3} // Float arithmetic -> 0.6667
{INT(CEILING(1.2)) / 3} // Integer arithmetic -> 0
```

In the current implementation, both will return 0. This PR should fix #409. My workaround ensures that unary operations return the same type as the argument, unless they are explicit casts (`INT()` or `FLOOR()`).

Let me know what you think!


ephread:
> Well it looks like at least a few month ago I was on board with this.

Yup! I based the workaround on your comment.

> I'm still a bit spooked that we end up converting too much

JavaScript shenanigans :/

PR #543, opened 2020-06-06
Using prettier

y-lohse:
This PR is meant to come after #519 which. is stuck in limbo right now.

Instead of searching and tweaking the eslint rules about formatting, I think we may as well use prettier rules and call. it a day. I ran the automatic fixing option so… there's a lot of changes, but it should be only cosmetics.


ephread:
Awesome, happy to use prettier!

PR #549, opened 2020-06-11
chore: Prettier everything

y-lohse:
Replaces #543


y-lohse:
I it was more the principle that needed to be reviewed than the result 😛

PR #556, opened 2020-06-12
fix: Remove calls to Array.push.apply

### LINKED ISSUE #176: Replace this.someArray.push.apply calls
There's a few `this.someArray.push.apply` lying around which are hard to read and can probably be written in some better form.

y-lohse:
## Checklist

## Description

Fixes #176 
The orignal code uses `Collection.AddRange` in several places, and using `Array.push.apply` was only one of the workarounds. There may be other places that could benefit from destructuring parameters passed to `Array.push`.


y-lohse:
Here we go again with the broken CI…

PR #558, opened 2020-06-12
chore: Switch to github actions

y-lohse:
Travis is failing to report it's status again for some obscure reason, there's no trace of errors anywhere and I can't be bothered re-installing the whole thing again. Github Actions presumably wokrs better on github..?

@ephread I think you mentioned using them before, could you let me know if I messed something up in the configuration?


.travis.yml
```diff
@@ -1,24 +0,0 @@
-language: node_js
-install:
-  - npm install -g codecov
```
y-lohse, line 3: @ephread why did we need the global install in travis only again? Should I add this line in the github actions, or could we specify it as a dev dep?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm run lint
+      - run: npm run build
+      - run: npm run test
+
+  publish-npm:
```
y-lohse, line 22: @ephread can you confirm that `test` and `publish-npm` are sequential and the second won't run if the first fails? This is what the doc seems to say but I had trouble finding a good explanation of this.

.travis.yml
ephread: I wouldn't put it in the dev dependencies, because we don't really need Codecov's uploader locally (we're never going to upload coverage). I'm not sure it needs to be installed globally though, I guess it was easier at the time!

.github/workflows/npm-publish.yml
ephread: Yes, steps are sequential. For instance, have a look at this [build log](https://github.com/ephread/Instructions/runs/658808761?check_suite_focus=true). The step named _Convert and upload coverage_ did not run because `Run tests` failed.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install
```
ephread, line 13: I would also name steps, but I guess npm <something> is already pretty straightforward.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,16 @@
+name: Merge check
```
ephread, line 1: Note that Github will use the name of the workflow for the default badge. So you'll get Merge check. It may not be what you want, but it can certainly be worked around by using shield.io.

y-lohse: Cool, I updated the names… and the badge that I forgot about.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on: pull_request
```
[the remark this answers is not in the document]
y-lohse: Mmh the master branch is protected from pushes, so I think it would not accomplish much, unless i misunderstood when `push` events happen.

.github/workflows/checks.yml
y-lohse: Ah I added job names, but I see you really can name steps. Let me add this quickly.

y-lohse:
So this is cool now but I need the travis check to go away please.

y-lohse:
So the commit removes the travis file and the travis integration is uninstalled so why exactly is that check still there..? Guess I'm creating a new PR to see if that works.

y-lohse:
Ok it was the github branch protection that was trying to enforce that outdated check.

PR #847, opened 2021-02-26
test: recompile tests with inklecate 1.0.0

ephread:
There isn't really anything to review in this PR, I'm just trying to keep the noise low for the next ones.


y-lohse:
Since I assume the tests are failing only due to the bumped ink version number, I wonder if we should also bump the version already, without doing the other changes in Story.ts, just in order to have the tests passing again?

ephread:
> I wonder if we should also bump the version already, without doing the other changes in Story.ts, just in order to have the tests passing again?

Very good idea, I hadn't thought of this! I'll make the change.

ephread:
It's updated. 93 out of the 238 tests are failing, but it's to be expected since the format changed.

y-lohse:
Yup cool, looks good! I'm happy to merge this into the V1.0 branch to track progress in that regard, if that was your intended plan?

y-lohse (approved):

ephread:
@y-lohse I originally wanted to merge all the test changes at once, but now that I think of it, having yet another intermediary branch is a bit overkill.

Happy to merge this in `v1.0` straight away!

PR #848, opened 2021-02-26
feat: Introduce BoolValue, migrate Value.ts

y-lohse:
Not sure this link actually works, but FWIW this is the diff https://github.com/inkle/ink/compare/0.9.0%E2%80%A6v1.0.0#diff-c987b0e17097c563f7313d33d30550eed37fd824af84a7e92ea8cc386d323146

### DIFF
--- a/src/engine/Value.ts
+++ b/src/engine/Value.ts
@@ -35,10 +35,8 @@ export abstract class AbstractValue extends InkObject {
       }
     }
 
-    // Implicitly convert bools into ints
     if (typeof val === "boolean") {
-      let b = !!val;
-      val = b ? 1 : 0;
+      return new BoolValue(Boolean(val));
     }
 
     // https://github.com/y-lohse/inkjs/issues/425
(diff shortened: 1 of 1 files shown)


src/engine/Value.ts
```diff
@@ -92,6 +90,44 @@ export abstract class Value<
...
+  }
+  public get isTruthy() {
+    return Boolean(this.value);
+  }
+  public get valueType() {
+    return ValueType.Int;
+  }
+
+  public Cast(newType: ValueType): Value<any> {
+    if (this.value === null) return throwNullException("Value.value");
```
y-lohse, line 105: I kept this in to stay consistent with the other casts, but:

1. It's not in the original source
2. It's not actually needed in this case because `null` gets evaluated to false

But I think we probably don't really want to consider null and false the same in this case.

ephread:
Great stuff, I'll give it a review tomorrow!

src/engine/Value.ts
```diff
@@ -92,6 +90,44 @@ export abstract class Value<
...
 
+export class BoolValue extends Value<boolean> {
+  constructor(val: boolean) {
+    super(val || false);
+  }
+  public get isTruthy() {
+    return Boolean(this.value);
+  }
+  public get valueType() {
+    return ValueType.Int;
```
ephread, line 101: Shouldn't this be `ValueType.Bool`?

src/engine/Value.ts
```diff
@@ -92,6 +90,44 @@ export abstract class Value<
...
+  }
+
+  public Cast(newType: ValueType): Value<any> {
```
ephread, line 104: Shouldn't the type be `Value<boolean>`?

src/engine/Value.ts
ephread: I honestly can't remember when this check was introduced and why!

Probably because if `this.value` is `null` in C#, the casts (at the Ink level, not at the language) would throw a null pointer exception while in Javascript, they would just go through silently.

It's probably trying to match the behaviour of the original runtime.

src/engine/Value.ts
y-lohse: Absolutely!

src/engine/Value.ts
y-lohse: Mmh no I don't think so, we're casting *from* boolean but to another type. To `newType` but I don't think we can use that in the type declaration. At best we could convert the `any` to `number | text`?

ephread: `any`'s good your right, I thought we were converting to booleans, sorry!

ephread (approved):

PR #850, opened 2021-02-27
test(v1.0): port new tests

ephread:
It should be a straightforward review!


y-lohse:
The branch has conflicts, looks like it has too many commits. But the actual changes look good! I didn't check the ones about flows in too much details, I think we'll know soon enough if everything in there is correct 😅

y-lohse (approved):

ephread:
> The branch has conflicts

@y-lohse what do you mean?

Screenshot 2021-03-05 at 16 08 12

> I didn't check the ones about flows in too much details, I think we'll know soon enough if everything in there is correct 😅

Haha yeah! We'll see if the test fails because of the code or because the test itself was ported incorrectly!

y-lohse:
I was looking at this

Capture d’écran 2021-03-07 à 18 42 24

But we can squash the commits too, I think it's ok :)

ephread:
> But we can squash the commits too, I think it's ok :)

Aaaaaah! Yeah for sure. I tend to squash them per PR anyway, it makes the history a bit cleaner without sacrificing much. I'll merge that one soon and start migrating some classes.

PR #855, opened 2021-03-12
feat: migrate DebugMetadata.ts

ephread:
I remember that code, I wrote it. 😎

It's not particularly needed by inkjs (it's a compiler thing), but I feel that if we don't migrate it, it'll cause some confusion down the line.


y-lohse (approved):
👍

PR #857, opened 2021-03-12
feat: migrate Flow.cs to Flow.ts

ephread:
This is a completely new class.

Nothing particularly stands out about that file.

### DIFF
--- a/src/engine/Flow.ts
+++ b/src/engine/Flow.ts
@@ -24,6 +24,7 @@ export class Flow {
     if (arguments[3]) {
       let jObject = arguments[3] as Record<string, any>;
 
+      this.callStack.SetJsonToken(jObject["callstack"], story);
       this.outputStream = JsonSerialisation.JArrayToRuntimeObjList(
         jObject["outputStream"]
       );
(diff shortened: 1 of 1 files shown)


src/engine/Flow.ts
```diff
@@ -0,0 +1,108 @@
...
+    let name = arguments[0] as string;
+    let story = arguments[1] as Story;
+
+    this.name = name;
+    this.callStack = new CallStack(story);
+
+    if (arguments[3]) {
+      let jObject = arguments[3] as Record<string, any>;
+
+      this.outputStream = JsonSerialisation.JArrayToRuntimeObjList(
```
y-lohse, line 27: I'm not 100% sure, but it seems like in this signature, we also need to do `this.callStack.SetJsonToken ((Dictionary < string, object > )jObject ["callstack"], story);` (l. 21 of Flow.cs https://github.com/inkle/ink/compare/0.9.0%E2%80%A6v1.0.0#diff-8cbd10a71465fea0e316355215ad5788ea62c3094e5b1e27d8e99eff1c1ee9f1R21). I think we may be missing that here?

y-lohse (commented):
One question, other than that LGTM!

src/engine/Flow.ts
ephread: Oh yeah, you're right, good catch! I'll fix that.

y-lohse (approved):
Sweet, thanks!

PR #858, opened 2021-03-12
feat: migrate Error.cs to InkError.ts

ephread:
I didn't migrate the delegate, since it's just take the form of a regular JavaScript function.

Easy review! 😅


y-lohse (approved):
Yep cool, although any reason to name the file InkError and not Error?

ephread:
@y-lohse No particular reasons, I created the file before really looking into what it entailed, I wanted to avoid a conflict with the native `Error` type, but this isn't going to happen. I'll change the name, it'll make more sense!

PR #859, opened 2021-03-12
feat: migrate NativeFunctionCall.ts

ephread:
This might be a bit annoying to review.

### DIFF
--- a/src/engine/NativeFunctionCall.ts
+++ b/src/engine/NativeFunctionCall.ts
@@ -4,7 +4,7 @@ import { Void } from "./Void";
 import { Path } from "./Path";
 import { InkList, InkListItem } from "./InkList";
 import { InkObject } from "./Object";
-import { asOrNull, asOrThrows } from "./TypeAssertion";
+import { asOrNull, asOrThrows, asBooleanOrThrows } from "./TypeAssertion";
 import { throwNullException } from "./NullException";
 
 type BinaryOp<T> = (left: T, right: T) => any;
@@ -208,7 +208,9 @@ export class NativeFunctionCall extends InkObject {
         return throwNullException(
           "NativeFunctionCall.CallBinaryListOperation op"
         );
-      let result = op(v1.isTruthy ? 1 : 0, v2.isTruthy ? 1 : 0);
+      let result = asBooleanOrThrows(
+        op(v1.isTruthy ? 1 : 0, v2.isTruthy ? 1 : 0)
+      );
       return new BoolValue(result);
     }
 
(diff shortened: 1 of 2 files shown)


src/engine/NativeFunctionCall.ts
```diff
@@ -209,7 +209,7 @@ export class NativeFunctionCall extends InkObject {
           "NativeFunctionCall.CallBinaryListOperation op"
         );
       let result = op(v1.isTruthy ? 1 : 0, v2.isTruthy ? 1 : 0);
```
y-lohse, line 211: I suggest we cast `result` with `Boolean` here, just to stay close to the source. I think BoolValue does a cast anyway, so not a huge deal.

src/engine/NativeFunctionCall.ts
```diff
@@ -428,34 +426,30 @@ export class NativeFunctionCall extends InkObject {
...
+      this.AddStringBinaryOp(this.NotEquals, (x, y) => !(x === y));
+      this.AddStringBinaryOp(this.Has, (x, y) => x.includes(y));
+      this.AddStringBinaryOp(this.Hasnt, (x, y) => x.includes(y));
 
       this.AddListBinaryOp(this.Add, (x, y) => x.Union(y));
       this.AddListBinaryOp(this.Subtract, (x, y) => x.Without(y));
-      this.AddListBinaryOp(this.Has, (x, y) => (x.Contains(y) ? 1 : 0));
-      this.AddListBinaryOp(this.Hasnt, (x, y) => (x.Contains(y) ? 0 : 1));
+      this.AddListBinaryOp(this.Has, (x, y) => x.Contains(y));
+      this.AddListBinaryOp(this.Hasnt, (x, y) => x.Contains(y));
```
y-lohse, line 437: I think we're missing a `!` here, eg. `!x.Contains(y)`

src/engine/NativeFunctionCall.ts
```diff
@@ -466,9 +460,8 @@ export class NativeFunctionCall extends InkObject {
       this.AddListUnaryOp(this.Count, (x) => x.Count);
       this.AddListUnaryOp(this.ValueOfList, (x) => x.maxItem.Value);
 
-      let divertTargetsEqual = (d1: Path, d2: Path) => (d1.Equals(d2) ? 1 : 0);
-      let divertTargetsNotEqual = (d1: Path, d2: Path) =>
-        d1.Equals(d2) ? 0 : 1;
+      let divertTargetsEqual = (d1: Path, d2: Path) => d1.Equals(d2);
+      let divertTargetsNotEqual = (d1: Path, d2: Path) => d1.Equals(d2);
```
y-lohse, line 464: Also missing the `!` here :)

y-lohse (changes requested):
Found 2 small errors in this one :)

I also wonder about using loose equality in most of this file rather than strict equality. I think it's ok because everything is already casted by the time we get here, so we should never end up comparing `1` and `"1"`… but I'm still a bit uncomfortable about that, now that I see it. Do you have any thoughts on that?

src/engine/NativeFunctionCall.ts
ephread: Wow, great catch! That would have been annoying to debug down the line. I'll fix that!

src/engine/NativeFunctionCall.ts
ephread: So `op` should already return a boolean value, since all the "ListBinaryOp" return a boolean. I kept it in line with what was there, but to get an equivalent behaviour, we would have to raise an exception if the return type isn't boolean rather than casting it the JavaScript way (it's a hard cast in C#). Happy to add it!

ephread:
> I also wonder about using loose equality in most of this file rather than strict equality. I think it's ok because everything is already casted by the time we get here, so we should never end up comparing 1 and "1"… but I'm still a bit uncomfortable about that, now that I see it. Do you have any thoughts on that?

I would keep comparisons as strict as possible to match the strong typing of C#, I feel that relying on JavaScript weak typing may cause problems down the line. That said, lets keep an eye on this and revisit once everything's ported, because maybe we'll uncover new issues.

src/engine/NativeFunctionCall.ts
y-lohse: Aaah right, I see. Nevermind, I think it's good as it is :)

y-lohse (approved):
All good, thanks!

src/engine/NativeFunctionCall.ts
ephread: Haha, I added the new `asBooleanOrThrows` type assertion 😅

PR #860, opened 2021-03-12
feat: migrate JsonSerialisation.ts

ephread:
I also disabled some warning outputted by the linter, because we can't fix them.


y-lohse (approved):
Looking good! 

Would you mind removing the `asNumberOrThrows` import since it appears to be unused?  Or is that the CI inventing things?

ephread:
> Would you mind removing the asNumberOrThrows import since it appears to be unused? Or is that the CI inventing things?

For sure, good catch, I'll update the PR and merge!

PR #861, opened 2021-03-12
docs: add CONTRIBUTING.md and start documenting the migration process

ephread:
I finally came around to create a `CONTRIBUTING.md`! I've been meaning to do this for a while, but I was lazy.

I plan to roughly formalise the migration rules (how to port certain structures), as well as the quirks we have in inkjs (such as testing against null to please TypeScript.

Let me know what you think!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,72 @@
+# Contributing to inkjs
+
+**⚠️ This document is a work in progress.**
+
+We're very happy to hear that you want to contribute! 🎊
+
+## Asking question / Reporting issues / Getting in touch
+
+### I have a question
+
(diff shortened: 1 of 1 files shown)


y-lohse:
Yep that's a great idea! I thought we already had something like this, but no.

What's there looks good to me. Off the top of my head, I think we should also mention our rules about comment (don't include comments from the original source, but add comments for js-specifics).

y-lohse (approved):

ephread:
> Off the top of my head, I think we should also mention our rules about comment (don't include comments from the original source, but add comments for js-specifics).

Good idea, let me add that to the doc immediately so that we don't forget.

y-lohse (approved):

PR #862, opened 2021-03-14
feat: migrate Story.ts

ephread:
There's a lot more changes than in the last few PR.

### Highlights

- The new Error Handler / Callbacks are types as `Function | null`, instead of just being optional, so that it matches the C# code better.
- There'a couple of errors, I'll fix them once `StoryState` is ported.
- The original integer overflow logic is different.
- I've tweaked the logic in `RemoveVariableObserver`, if I remember correctly a kind contributors created a PR in both repos, but the TypeScript code didn't seemt to match upstream. You may want to have a hard look at this.

### DIFF
--- a/src/engine/Story.ts
+++ b/src/engine/Story.ts
@@ -437,7 +437,7 @@ export class Story extends InkObject {
         this._state.variablesState.batchObservingVariableChanges = false;
 
       this._asyncContinueActive = false;
-      if (this.onDidContinue !== null) this.onDidContinue;
+      if (this.onDidContinue !== null) this.onDidContinue();
     }
 
     this._recursiveContinueCount--;
--- a/src/engine/Story.ts
+++ b/src/engine/Story.ts
@@ -437,7 +437,7 @@ export class Story extends InkObject {
         this._state.variablesState.batchObservingVariableChanges = false;
 
       this._asyncContinueActive = false;
-      if (this.onDidContinue != null) this.onDidContinue;
+      if (this.onDidContinue !== null) this.onDidContinue;
     }
 
     this._recursiveContinueCount--;
@@ -445,7 +445,7 @@ export class Story extends InkObject {
     if (this._profiler != null) this._profiler.PostContinue();
 
     if (this.state.hasError || this.state.hasWarning) {
-      if (this.onError != null) {
+      if (this.onError !== null) {
         if (this.state.hasError) {
           for (let err of this.state.currentErrors) {
             this.onError(err, ErrorType.Error);
--- a/src/engine/Story.ts
+++ b/src/engine/Story.ts
@@ -33,6 +33,7 @@ import { asOrNull, asOrThrows } from "./TypeAssertion";
 import { DebugMetadata } from "./DebugMetadata";
 import { throwNullException } from "./NullException";
 import { SimpleJson } from "./SimpleJson";
+import { ErrorHandler, ErrorType } from "./Error";
 
 export { InkList } from "./InkList";
 
@@ -87,6 +88,10 @@ export class Story extends InkObject {
     return this.state.currentWarnings;
   }
 
+  get currentFlowName() {
+    return this.state.currentFlowName;
+  }
+
   get hasError() {
     return this.state.hasError;
   }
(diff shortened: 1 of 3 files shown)


y-lohse:
Thank you, what a roll! I'll do my best but I may not be able to review this until much later in the week.

src/engine/Story.ts
```diff
@@ -387,16 +431,59 @@ export class Story extends InkObject {
       }
 
       this.state.didSafeExit = false;
+      this._sawLookaheadUnsafeFunctionAfterNewline = false;
 
       if (this._recursiveContinueCount == 1)
         this._state.variablesState.batchObservingVariableChanges = false;
 
       this._asyncContinueActive = false;
+      if (this.onDidContinue !== null) this.onDidContinue;
```
y-lohse, line 440: I think this should be calling onDidContinue, eg `this.onDidContinue()`

src/engine/Story.ts
```diff
@@ -387,16 +431,59 @@ export class Story extends InkObject {
...
+          sb.Append(this.state.currentErrors.length);
+          sb.Append(this.state.currentErrors.length == 1 ? " error" : "errors");
+          if (this.state.hasWarning) sb.Append(" and ");
+        }
+        if (this.state.hasWarning) {
+          sb.Append(this.state.currentWarnings.length);
+          sb.Append(
+            this.state.currentWarnings.length == 1 ? " warning" : "warnings"
+          );
+          if (this.state.hasWarning) sb.Append(" and ");
```
y-lohse, line 473: I think that's an extra line that isn't there upstream :)

y-lohse (commented):
I have to interrupt my review here, but I only have the RemoveVariableObserver left to check!

src/engine/Story.ts
ephread: Good catch!

src/engine/Story.ts
ephread: Haha bad copy/paste I suppose!

src/engine/Story.ts
```diff
@@ -1830,7 +1964,7 @@ export class Story extends InkObject {
   }
 
   public RemoveVariableObserver(
-    observer: Story.VariableObserver,
+    observer: Story.VariableObserver | null = null,
     specificVariableName: string
```
y-lohse, line 1968: Not sure this is part of the logic change, but I'll comment before forget: specificVariableName also defaults to null upstream.

y-lohse: Ah yes so it seems normal, but we can possibly type it as `specificVariableName?: string` to make it explicitly optional?

src/engine/Story.ts
```diff
@@ -1839,10 +1973,14 @@ export class Story extends InkObject {
...
-
         if (observer !== null) {
-          observers.splice(observers.indexOf(observer), 1);
+          let observers = this._variableObservers.get(specificVariableName)!;
+
+          if (observers !== null) {
+            observers.splice(observers.indexOf(observer), 1);
+          } else {
+            this._variableObservers.delete(specificVariableName);
+          }
```
y-lohse, line 1983: I think we may have a small difference in behavior here? Upstream, after removing the specific observer, if there are no more observers left for that variable, it is removed entirely (l. 2340 +2 in Story.cs).

src/engine/Story.ts
```diff
@@ -1852,7 +1990,12 @@ export class Story extends InkObject {
 
       for (let varName of keys) {
         let observers = this._variableObservers.get(varName)!;
-        observers.splice(observers.indexOf(observer), 1);
+
+        if (observers !== null) {
+          observers.splice(observers.indexOf(observer), 1);
+        } else {
+          this._variableObservers.delete(specificVariableName);
```
y-lohse, line 1997: Unless I'm missing something we have the same issue here as above: after removing the observer from this._variableObservers[varName], we don't call this._variableObservers.delete() even of that was the last of the observers.

y-lohse (commented):
Left some questions about the RemoveVariableObserver which may entirely be me misremembering something 😄

src/engine/Story.ts
ephread: I'd say… we can't a the moment, because we originally decided to closely match the C# "types", so `undefined` is avoided as much as possible.

But maybe we should reconsider now, and when it makes sense, use `undefined` instead of `null` to really highlight the optional nature of a parameter / property.

ephread: As for specificVariableName, I think what happened is that this code was originally written by another contributor and that rule isn't explicit.

Bu now that I think of it, it makes a of sense to stick to undefined for methods exposed outside of the runtime. Otherwise it's going to be confusing. I'll update that code and touch up the documentation at the same time.

src/engine/Story.ts
ephread: Yeah, you're right something's off here, let me update!

ephread:
I think I got it right this time!

y-lohse (approved):
Yep looks great, thanks!

PR #863, opened 2021-03-15
fix: fix mistakes preventing tests from passing

ephread:
This PR contains minor fixes. Tests are all green now thanks to @smwhr!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,9 @@ minified resulting JavaScript file.
 porting a new version of the runtime. The fixtures should always be compiled
 with the latest version of inklecate.
 
+⚠️ Running `test:javascript` and its variants won't transpile the files, so be
+sure to run `build` first.
+
 #### Porting tests from the reference implementation
 
 _To be added._
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,12 @@ the [main repository] first, or alternatively, in the official [Discord server].
    when it make sense. For example, it's important to document when `inkjs`
    display a slightly different behavior than the reference implementation.
 
+2. Try to match the C# type, that means that `String` translates to `string | null`
+   in TypeScript. In order to reduce confusion however, exposed methods
+   that interact with JavaScript should use optional parameter / return types
+   (i. e. `variable?: <type>`) when they are optional / nullable in
+   the original codebase.
+
 _To be added._
 
 ### Style guide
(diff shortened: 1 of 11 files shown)


smwhr:
In case this is useful, below is the mininal test that fails.
It does not even involve flow, just saving while having threads in there.  
Same test pass in the current public implementation.

```
it("tests simple save / load minimally with threads", () => {
    
    loadStory("multi_flow_save_load_threads");
    
    expect(story.Continue()).toBe("Default line 1\n");
    story.ChoosePathString("red");

    expect(story.Continue()).toBe("Hello I'm red\n");
    expect(story.currentChoices[0].text).toBe("Thread 1 red choice");

    let saved = story.state.ToJson();
    story.ResetState();
    story.state.LoadJson(saved);

    expect(story.currentText).toBe("Hello I'm red\n");
    expect(story.currentChoices[0].text).toBe("Thread 1 red choice");

    story.ChooseChoiceIndex(0);
    expect(story.ContinueMaximally()).toBe(
      "Thread 1 red choice\nAfter thread 1 choice (red)\n"
    );

  });
```

For the record, the exception is raised when "ContinueMaximally" is invoked and is the following :

```
StoryException: Ink had 1 warning and . It is strongly suggested that you assign an error handler 
to story.onError. The first issue was: RUNTIME WARNING: (Ink Pointer -> thread1.1.0.s -- index 2): 
Variable not found: 'name'. Using default value of 0 (false). This can happen with temporary 
variables if the declaration hasn't yet been hit. Globals are always given a default value on load 
if a value doesn't exist in the save state.
```

In order to trace which of the "name" variable was not found (because everything it that inkfile is called `name`), I tinkered with the ink file, and this is this one (that I renamed `flufty` to trace it manually):

```
== thread1(flufty) ==
+ Thread 1 {flufty} choice
    -> thread1Choice(flufty)
```

ephread:
Thanks to @smwhr the last test breaking the suite has been fixed 🎉. Thanks again for the fix!

Now we just need to look into why the JavaScript test fail, but it's likely to be a configuration issue.

ephread:
@smwhr my bad, I had forgotten to rebuild before running them. Gosh, it's been a while. I clarified that in `CONTRIBUTING.md`.

ephread:
@y-lohse you have the floor mate!

y-lohse (approved):
Yessss looking all good, thanks!

y-lohse:
This "all checks have passed" is really satisfying 😁

PR #879, opened 2021-05-16
chore: Run tests before publishing

y-lohse:
#878 worked but after seeing the workflow in action. I realize that both jobs run in parallel, but the publish step should run only if the test job is successful. So really the initial config was almost correct, except that `needs` should have been `test` and not `build`.


ephread (approved):
That's an easy review :P

PR #881, opened 2021-05-19
Typing changes for VariablesState and Story

evanwalsh:
## Checklist

## Description

This makes the experience of using inkjs from TypeScript slightly better by:

1. Adding typed access to a story's VariablesState. Now, a TypeScript user can do `story.variablesState['name']` or `story.variablesState.$<string>('name')` and either get the convenience that the indexed access affords _or_ the typed access via the generic param of `$`. Also, the `$` function didn't allow for fetching a value without a TypeScript error. That's fixed!
2.  In `Story.BindExternalFunction`, the `lookaheadSafe` parameter is now `false` by default. According to [Ink's source](https://github.com/inkle/ink/blob/0fe3c9f29d80a4eb4a9d7069796e357129f932ca/ink-engine-runtime/Story.cs#L1878-L1886):

    > Usually, you want to pass 'false', especially if you want some action to be performed in game code when this function is called

    As far as I can tell from the C#, `false` should generally be the default. This makes the specs' usage of `BindExternalFunction` correct and saves TypeScript users from having to pass `true` or `false` everytime!

### DIFF
--- a/src/tests/specs/ink/Bindings.spec.ts
+++ b/src/tests/specs/ink/Bindings.spec.ts
@@ -1,7 +1,8 @@
 import * as testsUtils from "../common";
+import { Story } from "../../../engine/Story";
 
 describe("Bindings", () => {
-  let story: any;
+  let story: Story | undefined;
 
   function loadStory(name: string) {
     story = testsUtils.loadInkFile(name, "bindings");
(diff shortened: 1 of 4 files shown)


y-lohse:
Hello, thank you for all 3 PRs! I've had a quick look already and they all seem interesting, I will try my best to do a review as soon as possible… but it might be a few days :)

src/tests/specs/ink/Bindings.spec.ts
```diff
@@ -16,15 +17,15 @@ describe("Bindings", () => {
 
     let testExternalBindingMessage = "";
 
-    story.BindExternalFunction("message", (arg: any) => {
+    story?.BindExternalFunction("message", (arg: any) => {
```
y-lohse, line 20: I'm guessing we are adding these to avoid a ts warning? If so I think it may be better to type story as `Story` only, and tell ts to ignore the line where we reset it to undefined in between tests?

y-lohse (commented):
Nice catch on the lookaheadSafe, we missed that in the last version :) Would you mind also changing the default value of `lookaheadSafe` for BindExternalFunctionGeneral to true? [We appear to have missed the default values for both.](https://github.com/inkle/ink/blame/master/ink-engine-runtime/Story.cs#L1887) If not no problem, I can do it in a separate PR.

Other than that LGTM, just a suggestion for one of the tests.

smwhr:
Next release : 
- Behaviour of `variablesState` has been modified in latest.
- Added default `true` to `lookaheadSafe`

PR #882, opened 2021-05-19
Switch `prepublishOnly` to `prepare` to allow for git installation

evanwalsh:
## Checklist

## Description

When trying to install the package from GitHub, I found that the built JavaScript files were missing! This adds that build step in by switching the `prepublishOnly` script to `prepare`.

Admittedly, this also means that running `npm install` in the repo's directory during development will always build the full JS. I'd like to make that build even faster 👹, but until then: `npm install --ignore-scripts` is the way to go.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "main": "dist/ink-es2015.js",
   "types": "ink.d.ts",
   "scripts": {
-    "test": "npm run-script test:typescript && npm run-script test:javascript",
+    "test": "npm run test:typescript && npm run test:javascript",
     "test:typescript": "jest",
-    "test:javascript": "npm run-script test:javascript:dist && npm run-script test:javascript:legacy",
+    "test:javascript": "npm run test:javascript:dist && npm run test:javascript:legacy",
     "test:javascript:dist": "INK_TEST=dist jest --config='jest.config.javascript.js'",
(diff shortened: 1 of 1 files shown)


y-lohse:
I think this was initially on purpose, since everything you would possibly do after installing the package locally had an automatic built step. But our dev steps have changed since and I don't see a good reason to *not* do a build after the install, so let's go with that :)

Thanks!

y-lohse (approved):

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "eslint 'src/**/*.ts'",
     "lint:fix": "npm run lint -- --fix",
     "watch": "tsc -p ./ && rollup -c -w",
-    "prepublishOnly": "tsc -p ./ && rollup -c"
+    "prepare": "npm run-script build"
```
y-lohse, line 18: Sorry just a quick change request: would you mind switching this to `npm run` instead of `npm run-script`, just to stay consistent with the existing commands?

y-lohse (changes requested):

package.json
evanwalsh: No problem! ✅ Done

y-lohse (approved):
Thanks!

PR #903, opened 2021-08-06
[request] add es6 module support

landongn:
## Goals
ES modules need a default export to function properly with other es modules.  Since this project is modern and uses modules under the hood, it was simple enough to add support by adding an additional export in the entry point along with the corresponding rollup build step. 

All tests pass, however there were some existing lint issues that remain from the moment of the fork to the modifications. 

Thanks!

### DIFF
--- a/ink.d.ts
+++ b/ink.d.ts
@@ -7,3 +7,4 @@ declare interface Inkjs {
 
 declare let inkjs: Inkjs
 export = inkjs
+export default inkjs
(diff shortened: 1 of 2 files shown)


landongn:
The reasoning behind adding support: using frameworks that operate entirely within modules have some problems upstream with vite and other watcher/compilers, and since there's no material change here I figured you may be interested in the updated support / compatability! thx

y-lohse:
Thank you! Yep cool, no problem with exposing this. We can probably remove the es-2015 version by now too.

PR #939, opened 2022-02-20
feat: port the remaining tests from Test.cs and fix minor issues

ephread:
## Checklist

## Description

I ported the remaining tests from upstream and also spend some time making sure ALL the tests were ported. It's a terrible PR, way too big to be reviewed comfortable. I apologize, it was hard to split up.

Out of the 107 files, all the `*.ink` (37) and `*.json` (17) can be ignored. All the specs under inkjs can also be ignored, they are noise. I just updated them to use the `TestContext`, they still pass happily.

Only eight tests are failing:

- _Variables › tests const redefinition_
- _Variables › tests variable naming collision with arg_
- _Choices › tests various blank choice warning_
- _Functions › tests argument name collision_
- _Functions › tests function purity checks_
- _Knots › tests stitch naming collision_
- _Misc › tests comment eliminator_
- _Misc › tests return text warning_

They're all compiler-related and it might be useful to review them thoroughly. I may have made a mistake while porting them. The name of the original test can be found as a comment above each case.

I've self-reviewed to provide some context to the PR, but it's probably not worth spending too much time trying to get everything right. I tried to be quick so that we can move forward with the compiler. We'll refactor the test infrastructure later (and boy, it needs a fresh coat of paint – it wasn't designed to work with a JS compiler 😅).

One last thing, there are two more tasks remaining:

1. ensure the test pass on the dist files (`test:javascript` fails at the moment);
2. test that inklecate's JSON is the same as inklecatejs's JSON.

Feel free to raise any issue and discuss the comments below.


src/compiler/FileHandler/PosixFileHandler.ts
```diff
@@ -2,14 +2,17 @@ import { IFileHandler } from "../IFileHandler";
 import * as path from "path";
 import * as fs from "fs";
 
+// This class replaces upstream's DefaultFileHandler.
 export class PosixFileHandler implements IFileHandler {
```
ephread, line 6: I made some changes to POSIX file handler, because I needed a file handler that was closer to `DefaultFileHandler` to be able to run some of the tests (the ones involving 'INCLUDE').

Also, see `InkParse.ts` below, no default file handler was registered. I'm not certain what was the original intention, maybe eventually assign a different handler by default, depending on the execution environment?

src/compiler/Parser/ErrorType.ts
```diff
@@ -1,5 +1,7 @@
+// TODO: Unifify with Engine.
```
ephread, line 1: `Unifify`, MY BAD. I'll fix it in a subsequent PR.

src/compiler/Parser/StringParser/StringParser.ts
```diff
@@ -16,7 +16,7 @@ export type ParseRuleReturn =
 
 export type SpecificParseRule<T extends ParseRule> = T;
 
-export abstract class StringParser {
+export class StringParser {
```
ephread, line 19: I removed `abstract` because some tests use `StringParser` as a concrete class and it's (seemingly) not abstract in upstream.

src/tests/compile.js
```diff
@@ -1,68 +1,114 @@
...
+  if (errors.length === 0) {
+    console.log("Done.");
+  } else {
+    errors.forEach((error) => console.error(`\n${error.reason.message}`));
+  }
+}
 
-	let results = await Promise.allSettled(promises);
-	let errors = results.filter(result => result.status === "rejected");
+let inklecate = process.argv[3];
```
ephread, line 108: Since the tests are autocompiling now, this file is back to using the original inklecate. An improvement: it can now be provided, instead of relying on its presence in $PATH. We'll use the json stored in the repo to perform a diff, but that'll come in a future PR.

src/tests/specs/common.ts
```diff
@@ -1,27 +1,55 @@
 import * as path from "path";
```
ephread, line 1: This file was heavily rewritten, but it's now a mess. I focused on getting the tests out he door, the test infrastructure is bound to get refactored soon. I'll open an issue.

src/tests/specs/common.ts
```diff
@@ -50,3 +87,83 @@ function getRootDir() {
     return path.join(__dirname, "..", "..", "..");
   }
 }
+
+export class TestContext {
```
ephread, line 91: `TestContext` is an object that stores a bunch of values across all the tests of a given spec. It's useful to reduce the amount of boilerplate. Not super happy about it for now.

src/tests/specs/inkjs/compiler/Core.spec.ts
```diff
@@ -24,10 +23,10 @@ describe("Core parsers", () => {
     expect(parser.index).toBe(24);
   });
 
-  it("parses newLine", () => {
-    const parser = new InkParser(`moo text and 
-then an -> happens
-and what ?`);
+  it.only("parses newLine", () => {
+    const parser = new InkParser(
+      `moo text and \nthen an -> happens\nand what ?`
```
ephread, line 28: I turned those into `\n`-bound strings because my VS Code removes trailing spaces on save. Prettier is a lot smarter and can detect that trailing spaces should be kept in a template string, but to prevent future headaches I made the whitespace more explicit.

src/tests/specs/setupTests.ts
```diff
@@ -0,0 +1,39 @@
...
+declare global {
+  namespace jest {
+    interface Matchers<R> {
+      toContainStringContaining(expected: string): CustomMatcherResult;
+    }
+  }
+}
+
+expect.extend({
+  toContainStringContaining(
```
ephread, line 12: Useful matcher to check whether the error / warning arrays contained the offending messages.

src/compiler/FileHandler/PosixFileHandler.ts
smwhr: When packaging for browser use (the main use case imo), it did not really make sense to assign the Posix/Filesystem handler, this is the main reason I named it that way (Default would not actually be default !).
Moreover, it creates a dependancy to node-fs that is in turn not resolvable for the web package.

src/tests/specs/inkjs/compiler/Core.spec.ts
smwhr: Works for me. Templates with new lines were mainly a side effect of me copy-pasting the examples around, this is indeed cleaner that way !

src/compiler/Parser/InkParser.ts
```diff
@@ -110,6 +115,7 @@ export class InkParser extends StringParser {
   }
 
   // Main entry point
+  // NOTE: This method is named Parse() in upstream.
```
smwhr, line 118: FYI : Naming it Parse broke the inheritance (StringParse.Parse has nothing to do with this method)

src/tests/specs/inkjs/compiler/Core.spec.ts
ephread: Happy to hear that!

src/compiler/FileHandler/PosixFileHandler.ts
ephread: Makes sense! Do you think there's a way we could have a different default handler depending on the environment in InkParser?

I hope we can keep a default handler, otherwise, we need to flag this as an important difference and document it accordingly.

smwhr:
Fixed 2 of them (and I discovered that redefining a CONST is perfectly legal if you do not change the value)

smwhr:
@ephread I just PR-ed your branch on your repo https://github.com/ephread/inkjs/pull/2#pullrequestreview-904015216, that fixes all the remaining tests !

src/compiler/FileHandler/PosixFileHandler.ts
smwhr: Which tests were failing due to lack of default fileHandler ? Looks like you fixed it because when I comment it out it still works.

I think documentation is the way to go here, because even in a browser, it's not clear which "FileHandler" the user wants to use and I think it's a good idea to make people stop and think about how they want to handle multiple files (and if they don't it'll work without default anyway)

ephread: > Which tests were failing due to lack of default fileHandler ? Looks like you fixed it because when I comment it out it still works.

[The tests explicitly use PosixFileHandler](https://github.com/y-lohse/inkjs/pull/939/files#diff-d3f98f7c5f5e3b3c64662aef9400b1310a81b137aaf1e7c3474bafe05467ec67R133) 😇 

> I think documentation is the way to go here, because even in a browser, it's not clear which "FileHandler" the user wants to use and I think it's a good idea to make people stop and think about how they want to handle multiple files (and if they don't it'll work without default anyway)

Alright, we need to replace that by a check and an explicit error then. Let me update the PR!

ephread:
@smwhr awesome, thanks! Let me have a look.

ephread:
@smwhr there's quite a bit of change in https://github.com/ephread/inkjs/pull/2#pullrequestreview-904015216, I feel like we should merge this PR first and then recreate your PR in inkjs rather than my fork. Reviews will be hard to manage otherwise.

Let me update the PR and we should be able to merge it.

ephread:
[66e7178](https://github.com/y-lohse/inkjs/pull/939/commits/66e717812c1d486af516da7f093d78d6c3fb9da3) solves the issue raised in [this comment](https://github.com/y-lohse/inkjs/pull/939#discussion_r811471534). I think it's elegant, but let me know what you think.

In short, the default file handler throws an exception when resolving / loading. If the entire Ink story is provided as a String, there's no need to provide a handler, but otherwise, the error is more explicit than just throwing something like "FileHandler is null". We also keep the source code close to upstream in InkParser.

smwhr:
> [66e7178](https://github.com/y-lohse/inkjs/pull/939/commits/66e717812c1d486af516da7f093d78d6c3fb9da3) solves the issue raised in [this comment](https://github.com/y-lohse/inkjs/pull/939#discussion_r811471534). I think it's elegant, but let me know what you think.

It is indeed elegant and the error message being explicit will make a great Google query that points to an issue here where the clueless developper will find answers to their answers. 👍🏼

ephread:
Awesome! I'll merge this, I think we resolved all the discussions. Then you can recreate your PR against `compiler`. No need to port over the comments from the other PR, I'll switch between the two PR during the review.

PR #944, opened 2022-03-13
Compiler tests : the last mile

smwhr:
## Checklist

## Description
- Fixed all ts tests
- Fixed all js tests
- Added typeName to all Elements for easier debugging


smwhr:
Original comments here https://github.com/ephread/inkjs/pull/2

ephread:
Starting the review!

src/compiler/Parser/ParsedHierarchy/Declaration/ConstantDeclaration.ts
```diff
@@ -32,6 +32,10 @@ export class ConstantDeclaration extends ParsedObject {
     }
   }
 
+  get typeName(): string {
+    return "CONST";
```
ephread, line 36: It's different from upstream, but it shouldn't make any difference I guess? Any reason for the change?

src/compiler/Parser/ParsedHierarchy/Declaration/ExternalDeclaration.ts
```diff
@@ -15,6 +15,10 @@ export class ExternalDeclaration extends ParsedObject implements INamedContent {
     super();
   }
 
+  get typeName(): string {
+    return "EXTERNAL";
```
ephread, line 19: Thinking about it some more, I guess it looks better that way that `GetType().Name` from upstream. 👍

We should document those differences, in case someone else looks at the code down the line. It's a bit tedious, but… it's helpful in the long run!

src/compiler/Parser/ParsedHierarchy/Expression/NumberExpression.ts
```diff
@@ -1,6 +1,8 @@
 import { Container as RuntimeContainer } from "../../../../engine/Container";
 import { Expression } from "./Expression";
 import { BoolValue, FloatValue, IntValue } from "../../../../engine/Value";
+import { asOrNull } from "../../../../engine/TypeAssertion";
+import { ParsedObject } from "../Object";
 
 export class NumberExpression extends Expression {
```
ephread, line 7: Comment for later, it's named `Number` in upstream, even though this name makes more sense to me. To be documented later on.

src/compiler/Parser/ParsedHierarchy/Expression/NumberExpression.ts
```diff
@@ -39,4 +45,14 @@ export class NumberExpression extends Expression {
   };
 
   public readonly toString = (): string => String(this.value);
+
+  public Equals(obj: ParsedObject): boolean {
+    const numberExpression = asOrNull(obj, NumberExpression);
+    if (!numberExpression) return false;
+
+    return (
+      numberExpression.subtype == this.subtype &&
```
ephread, line 54: Smart 👍

src/compiler/Parser/ParsedHierarchy/Variable/VariableAssignment.ts
```diff
@@ -28,6 +28,9 @@ export class VariableAssignment extends ParsedObject {
     if (this.isNewTemporaryDeclaration) {
       return "temp";
     } else if (this.isGlobalDeclaration) {
+      if (this.listDefinition !== null) {
```
ephread, line 31: That's better! To be documented as well.

src/tests/specs/ink/Variables.spec.ts
```diff
@@ -208,9 +207,8 @@ describe("Variables", () => {
     // The Original code used 'CompileString', but since the compilation fails,
     // 'compile' can be used instead.
     compileStoryWithoutRuntime("variable_naming_collision_with_arg");
-
     expect(context.errorMessages).toContainStringContaining(
-      "name has already been used for a function"
+      "name has already been used for a argument to knot"
```
ephread, line 211: That's a good catch, upstream simply looks for `"has already been used"`, which I assume is better. I wasn't fully awake when shuffling code around.

src/tests/specs/ink/Misc.spec.ts
```diff
@@ -196,8 +196,7 @@ C C C
   // TestReturnTextWarning
   it("tests return text warning", () => {
     compileStoryWithoutRuntime("return_text_warning");
-
-    expect(context.warningMessages.length).toBe(1);
+    expect(context.warningMessages.length).toBe(2);
```
ephread, line 199: Well, seems that upstream doesn't check the number of warnings, so I guess `toBeGreaterThanOrEqual(1)` would work just as well!

src/tests/specs/ink/Evaluation.spec.ts
```diff
@@ -23,7 +23,9 @@ describe("Evaluation", () => {
   // TestArithmetic
   it("tests arithmetic", () => {
     compileStory("arithmetic");
-    expect(context.story.ContinueMaximally()).toBe("36\n2\n3\n2\n2\n8\n8\n");
+    expect(context.story.ContinueMaximally()).toBe(
```
ephread, line 26: I remember now why I paused the review the first time, I couldn't understand why inkjs was again able to differentiate between `1.0` & `1` … 🤦🏻‍♂️

Now I see it, it's because it doesn't go through the JSON intermediate representation here. We need to remember to revisit this test soon, so that we can assess the behaviour when loading a JSON file.

src/tests/specs/ink/Extra.spec.ts
```diff
@@ -24,7 +24,7 @@ describe("Extra", () => {
     compileStory("arithmetic_2");
 
     expect(context.story.ContinueMaximally()).toBe(
-      "2\n2\n2.3333333333333335\n2\n2\n2.3333333333333335\n"
+      "2\n2.3333333333333335\n2.3333333333333335\n2.3333333333333335\n2.3333333333333335\n2.3333333333333335\n"
```
ephread, line 27: Same here!

ephread (approved):
LGTM!

I've added a bunch of comments, so we won't forget to have another look at this later, but otherwise it's good. Thank you so much for fixing all the tests and the typos and sorry it took so long to review. Life happened. 😅

It's been in limbo for too long, let's merge this!

src/compiler/Parser/ParsedHierarchy/Expression/NumberExpression.ts
smwhr: [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) is a predefined class in js hence the new name. I'll add a comment in code.

src/compiler/Parser/ParsedHierarchy/Declaration/ConstantDeclaration.ts
smwhr: It weirded me out that the type did not look like its declaration (it's the case for the ones that were already defined) and created a special exception when playing with the parsedHierrarchy tree.

src/compiler/Parser/ParsedHierarchy/Expression/NumberExpression.ts
ephread: Works for me!

Issue #935, opened 2022-02-18
[Compiler] What name should we use for the JavaScript version of inklecate?


ephread:
Per #933, the JavaScript version of inklecate is likely to have a slightly different interface compared to the original implementation. To avoid confusion, I think we should name the command differently.

Before running a poll, I'd like the input of the community. We currently have a short list of names, and new options are welcome! I'll update this issue with any suggestions made either here or on Discord.

**——— EDIT ———**

Here are all the suggestions made by the community:

1. inklenode
2. inklet
3. inklescript
4. nodlecate
5. inkjs-compiler
6. inklecatejs
7. inkycatjswriter (😂 – Not part of the poll!)

I would also like to suggest another name 🙃 (NO SPOILERS!):

![Screen Shot 2022-02-20 at 20 59 47](https://user-images.githubusercontent.com/722366/154862109-8e8e8cd5-e2db-4e82-97af-a47bedc1524c.png)

## Poll

**Vote now!** – _Click on the option you prefer the most._

[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklenode)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklenode/vote)
[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklet)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklet/vote)
[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklescript)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklescript/vote)
[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/nodlecate)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/nodlecate/vote)
[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inkjs-compiler)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inkjs-compiler/vote)
[![](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklecatejs)](https://api.gh-polls.com/poll/01FWCBPNXGYRQRFD4652KW45NW/inklecatejs/vote)
[...]

PR #954, opened 2022-05-07
Rename inklecate to inkjs-compiler

smwhr:
The community [has voted](https://github.com/y-lohse/inkjs/issues/935), so the renaming shall be !

## Checklist

## Description

Rename inklecate to inkjs-compiler
There was a small bug with `INCLUDE`s when using `inkjs-compiler` (path resolution was wrong), it's been updated.


src/tests/compile.js
```diff
@@ -101,11 +104,11 @@ async function compileInkFile(inklecate) {
   if (errors.length === 0) {
     console.log("Done.");
   } else {
-    errors.forEach((error) => console.error(`\n${error.reason.message}`));
+    errors.forEach((error) => console.error(`\n${error.reason.message.substring(0,255)}`));
```
smwhr, line 107: When using the minified "inkjs-compiler.js", error message include the full unreadable stack trace. 255 chars should be enough to identify which test failed and run it individually.

src/tests/compile.js
```diff
@@ -101,11 +104,11 @@ async function compileInkFile(inklecate) {
...
   }
 }
 
-let inklecate = process.argv[3];
+let inklecate = process.argv[2];
```
smwhr, line 111: `[2]` was not used, not sure why it looked at `[3]`

PR #993, opened 2022-12-12
Fix INCLUDE paths in inkjs-compiler

iainmerrick:
## Checklist

## Description

This fixes a couple of problems I ran into when trying to run `inkjs-compiler` on Ink source split across multiple files:

- Adds a basic error handler that prints errors to process.stderr
- Don't mangle filenames in `PosixFileHandler`

The file handler was removing the current directory from anywhere in the filename, which is definitely incorrect (e.g. if the current directory is "." it will remove all dots from the filename).

I guess the intention may have been to the remove the current directory from the start of the filename, but I don't think that's correct either — it would break if the working directory were say `ink` and we tried to include `ink/ink/foo.ink`.

So, this PR just interprets the filename as being relative to the working dir, which does the right thing for my use case.

### DIFF
--- a/script/inkjs-compiler.ts
+++ b/script/inkjs-compiler.ts
@@ -6,6 +6,7 @@ var readline = require('readline');
 var path = require('path');
 
 import * as fs from "fs";
+import { ErrorHandler } from '../src/engine/Error';
 
 const help = process.argv.includes("-h");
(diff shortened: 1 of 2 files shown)


iainmerrick:
Thanks for the review!

PR #997, opened 2023-02-20
Fix incorrect substring in TrySplittingHeadTailWhitespace

iainmerrick:
## Checklist

## Description

This method was calling [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) with (start, length) instead of (start, end).

I ran into this with an ink file including command output that looks like this: `\nFoo bar`. It would print `Foo ba` instead of `Foo bar`.

I'm not an ink expert, so I could use some help adding a unit test for this!

### DIFF
--- a/src/engine/StoryState.ts
+++ b/src/engine/StoryState.ts
@@ -815,10 +815,7 @@ export class StoryState {
     }
 
     if (innerStrEnd > innerStrStart) {
-      let innerStrText = str.substring(
-        innerStrStart,
-        innerStrEnd - innerStrStart
-      );
+      let innerStrText = str.substring(innerStrStart, innerStrEnd);
       listTexts.push(new StringValue(innerStrText));
     }
 


smwhr:
Wow, how did we miss that ?!

Do you have a reproducible minimal ink snippet to share ?
We can add it to the test suite quite easily.

smwhr:
(And I guess the same mistake exists at line 830 of the same file)

iainmerrick:
Sorry, I don't have a minimal case, it's buried deep inside a big game!

smwhr:
No problem :) 
I'll try and see how to trigger this bug and merge this during the weekend !

smwhr:
Found it !

smwhr:
@iainmerrick thanks for the fix ! It is now released.

PR #1002, opened 2023-04-06
simplify POSIX file handler

zekecato:
## Checklist

## Description

Initialize rootPath with a default value (empty string), and use path.resolve() to construct an absolute path. If an absolute path is provided, that will be used. if a relative path is provided, it will append that to the working directory.


smwhr:
LGTM if anyone else want to second the change.

PR #1036, opened 2023-11-23
Fixed maxInt type check in RANDOM

floriancargoet:
Probably a copy-paste error when duplicating the code from `minInt` (a few lines above) for `maxInt`.
The instanceof check should be on maxInt.


smwhr:
Thank you !

Issue #959, opened 2022-05-20
Typescript error about export assignment


SpoonMeiser:
## Describe the bug

I've just updated a typescript project to inkjs 2.1.0, and when I use typescript to check the types, it throws an error:

```
$ npx tsc --noEmit
node_modules/inkjs/ink.d.ts:13:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.

13 export = inkjs
   ~~~~~~~~~~~~~~
```

Checking the file, I see:

```
export = inkjs
export default inkjs
```

Which seems to be unsupported according to https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

## To Reproduce

Have a typescript project that imports ink, and run `npx tsc --noEmit`

## Expected behavior

`npx tsc --noEmit` to return no errors

## Ink files

N/A

## Environment
 - OS: Ubuntu
 - Node.js version: v16.14.2
 - Web browser version: N/A
 - _inklecate_ version: N/A
 - _inkjs_ version: 2.1.0

SpoonMeiser:
Some additional information

I'm importing ink with:

```
import { Story as InkStory } from "inkjs"
```

and using typescript version: 4.6.3

PR #1049, opened 2024-03-31
Remove export assignment from ink.d.ts

maetl:
## Description

As described in #959, there is an incompatibility between how this declaration file is structured and the expectations of recent TypeScript versions.

Removing the export assignment ensures that only the default symbol is exported.

```
An export assignment cannot be used in a module with other exported elements.
```

Perhaps a better long term solution would be to upgrade the version of TypeScript used in the project. Happy to help with this if I can get a bit of guidance around where to start and managing build tools.

## Checklist

### DIFF
--- a/ink.d.ts
+++ b/ink.d.ts
@@ -1,8 +1,8 @@
-import { Story, InkList } from './engine/Story'
-import { Compiler } from './compiler/Compiler'
-import { CompilerOptions } from './compiler/CompilerOptions'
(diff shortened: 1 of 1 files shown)


smwhr:
Merged.

Ask away if you need guidance/info !

maetl:
Merci! Confirmed working in the typescript project where I first noticed the issue.

I’ll ask additional questions in a separate issue thread.

PR #1052, opened 2024-04-06
Fix imports and add docblocks

maetl:
## Description

Fixes a problem introduced in #1049 where the relative imports changed to suit standalone references to these files on disk—but overlooked that the tsconfig had `src` already mapped.

## Checklist


smwhr:
I was wondering why those imports were changed indeed in your last PR but did not took the time to check.

It seems the namespace is now Capitalized, whereas is was all lowercase before. It's important for js includes.

maetl:
> It seems the namespace is now Capitalized, whereas is was all lowercase before. It's important for js includes.

Fixed now. Thanks for explaining.

PR #1089, opened 2024-11-16
Added stats generation

brwarner:
## Checklist

- [X] The new code additions passed the tests (`npm test`).
- [X] The linter ran and found no issues (`npm run-script lint`).
- [x] Check the ink repository for any tests to migrate surrounding stats

## Description

Adding stats generation from the ink compiler to inkjs. The ink compiler only exposes this as a command line option, but for inkjs I've exposed it as a new function in the Compiler that can be run after compilation.

For source from the ink repository, see https://github.com/inkle/ink/blob/master/compiler/Stats.cs

### DIFF
--- a/src/compiler/Compiler.ts
+++ b/src/compiler/Compiler.ts
@@ -7,6 +7,7 @@ import { Story as ParsedStory } from "./Parser/ParsedHierarchy/Story";
 import { DebugMetadata } from "../engine/DebugMetadata";
 import { StringValue } from "../engine/Value";
 import { asOrNull } from "../engine/TypeAssertion";
+import { GenerateStoryStats, Stats } from "./Stats";
 
 export { CompilerOptions } from "./CompilerOptions";
 export { InkParser } from "./Parser/InkParser";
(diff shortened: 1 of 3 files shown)


brwarner:
There are no tests in the ink repository around story stats

smwhr:
Thanks for porting this stat counter !

Although adding one simple test case for the function to insure no one breaks it unintentionally in the future would be a nice addition, don't you think ?

brwarner:
For sure, I'll add one in!

brwarner:
Tests are in `src/tests/specs/inkjs/compiler/Stats.spec.ts`

smwhr:
Merged ! 🚀

smwhr:
The number of diverts is not counted in the same manner as when called with the original inklecate.

Ex, on the file `src/tests/inkfiles/original/inkjs/tests.ink`
- this implementations reports 101 diverts
- the inklecate implementation reports 131 diverts

This has nothing to do with your code, so I will just open a new issue for later resolution.
Also, who is really interested in the number of diverts ?

PR #1100, opened 2025-02-08
Add aggressiveFloatParsing

smwhr:
## Checklist

- [ ] The linter ran and found no issues (`npm run-script lint`).

## Description

Add `aggressive float parsing` option so that when loading a json file that contain explicit floating point number ending in `.0`, they are understood as `FloatValue` and not `IntValue`, leading to errors in games loaded with previous version of inkjs.

### DIFF
--- a/src/engine/SimpleJson.ts
+++ b/src/engine/SimpleJson.ts
@@ -1,17 +1,34 @@
 export class SimpleJson {
-  public static TextToDictionary(text: string) {
-    return new SimpleJson.Reader(text).ToDictionary();
+  public static TextToDictionary(
+    text: string,
+    aggressiveFloatParsing: boolean = false
+  ) {
+    return new SimpleJson.Reader(text, aggressiveFloatParsing).ToDictionary();
   }
 
-  public static TextToArray(text: string) {
-    return new SimpleJson.Reader(text).ToArray();
+  public static TextToArray(
+    text: string,
+    aggressiveFloatParsing: boolean = false
+  ) {
+    return new SimpleJson.Reader(text, aggressiveFloatParsing).ToArray();
   }
 }
 
 export namespace SimpleJson {
   export class Reader {
-    constructor(text: string) {
-      this._rootObject = JSON.parse(text);
+    constructor(text: string, aggressiveFloatParsing: boolean = false) {
+      if (aggressiveFloatParsing) {
+        // When the aggressiveFloatParsing argument is true, we aggressively replace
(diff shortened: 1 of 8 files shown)


floriancargoet:
`JSON.parse` can take a `reviver` function as second parameter. On modern browsers, this reviver [has access to the source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#context) of the value being parsed.

```js
JSON.parse('[3.0]', (key, value, context) => {
  // key is 0, value is 3, context is { source: "3.0" } 
  return value;
})
```

With that, we could convert 3.0 to "3.0f" without being aggressive. Something like: 
```js
JSON.parse('{"key":"value", "array": [1, 2, null, 3.0, false]}', (key, value, context) => {
  if (Number.isInteger(value) && context.source.endsWith(".0")) {
    return context.source + "f";
  }
  return value;
})
```

And it's not hard to detect support for context: 
```js
function testHasReviverContext() {
  return JSON.parse("0", (key, value, context) => context != null)
}
```

smwhr:
_That_ is very interesting !

smwhr:
But safari (osx and ios) doesn't support it, that's unfortunate

floriancargoet:
Yes, that's why I mentioned testing the support.
It could be used where it works by default and there's the aggressive option where it doesn't.

src/engine/SimpleJson.ts
```diff
@@ -1,33 +1,49 @@
 export class SimpleJson {
   public static TextToDictionary(
-    text: string,
-    aggressiveFloatParsing: boolean = false
```
floriancargoet: Don't you want to keep the option around for Safari?

smwhr: I chose the lesser of two evils :
- either the math is totally wrong on safari unless aggressiveFloatParsing is explicit
- or there's a very minuscule chance of ", 5.0" being displayed ", 5.0f" in a text somewhere

smwhr: By essence, you never know where your code will be executed (Safari or elsewhere) so you basically never have the choice to turn the aggressiveFloatParsing. It is now on by default for Safari. And for everyone else, I use your suggestion.

floriancargoet: I read to fast and missed the fact that you only removed the option but kept the aggressive code. 👍

floriancargoet:
I think this part of the documentation can be removed or edited:
https://github.com/y-lohse/inkjs/blob/master/docs/compiler-differences.md#float-and-ints

PR #1116, opened 2025-06-22
feat: compiler: add source filename to the IFileHandler

bemisguided:
## Checklist

## Description

Add a `sourceFilename` parameter to the `ResolveInkFilename` and `LoadInkFileContents` methods of the `IFileHandler`. These will provide context to the IFileHandler to be able to resolve the actual Include Filename, relative to the Source Ink Filename that included it. This is particularly useful when an Ink Story has Include Files that, in themselves, also Include Files.

For example, `/main.ink` includes `/common/functions.ink`, which includes `/common/library/my_function.ink`, this would open the door to define as follows:

```/main.ink
INCLUDE common/functions.ink
```

```/common/functions.ink
INCLUDE library/my_functions.ink
```

This PR is being proposed in support of a [VSCode Extension for Ink](https://github.com/bemisguided/vscode-ink-language-tools) under-development.

On Discord, we discussed concerns about breaking compatibility with Inky, which this most certainly open the door for. It was agreed that the VSCode Extension should offer an "Inky compatibility" setting (default to enabled), to avoid confusion, when using this new feature.

### DIFF
--- a/src/compiler/IFileHandler.ts
+++ b/src/compiler/IFileHandler.ts
@@ -1,6 +1,12 @@
 export interface IFileHandler {
-  readonly ResolveInkFilename: (filename: string) => string;
-  readonly LoadInkFileContents: (filename: string) => string;
+  readonly ResolveInkFilename: (
+    filename: string,
+    sourceFilename?: string | null
+  ) => string;
+  readonly LoadInkFileContents: (
+    filename: string,
+    sourceFilename?: string | null
+  ) => string;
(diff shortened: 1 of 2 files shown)


bemisguided:
I did not see any obvious places to update tests for this. I'd be happy to add tests, if I missed the places, or it was felt we should add some.

Further, I suspect the README should be updated with this change. I'd also be happy to update this PR with this as well.

